PR #4, opened 2018-06-26
More tests

dkfellows:
Check the remaining functionality in this branch


Christian-B:
As the Laird would say "Rubbered"

PR #5, opened 2018-06-26
Java package structure

dkfellows:
Build out more of the metadata structure

The `RemoveMe` classes are just there to stop the javadoc build from blowing up, and should be deleted once we have real code in those packages.


rowleya:
Just a note - spinnaker.manchester.ac.uk doesn't exist; spinnaker.cs.manchester.ac.uk does.

rowleya:
Another note - is SpiNNaker-transceiver what you want as the name?  Would SpiNNaker-comms be better (agree that SpiNNMan probably doesn't say much).  Similarly, we could take this opportunity to make SpiNNaker-PACMAN into SpiNNaker-mapping?

dkfellows:
Re the overall name (and whether it has `cs` in it), that's only important if another project or organisation in the university is called “spinnaker” _and_ code from there is used in the same process. Vanishingly unlikely. Otherwise it just adds noise.

dkfellows:
Maven submodules can be renamed easily at this stage as we've not got lots of inter-module dependencies yet. If we know what they should be, I'll just change them.

PR #14, opened 2018-08-21
Progress

Christian-B:
Adds ProgressBar and Iterator.
Added time to ProgressBar.

- [ y] javadoc correct.
- [ y] test case(s).


SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/UnitConstants.java
```diff
@@ -20,4 +25,40 @@ private UnitConstants() { }
...
+            Date date = new Date(durationInMillis);
+            return formatter.format(date) + " s";
+        }
+        if (durationInMillis < MS_PER_H) {
+            DateFormat formatter = new SimpleDateFormat("m:ss.SSS");
+            formatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+            Date date = new Date(durationInMillis);
+            return formatter.format(date) + " m";
+        }
+        DateFormat formatter = new SimpleDateFormat("H:mm:ss.SSS");
```
dkfellows: This will be _fun_ once we deal with durations longer than 24 hours.

SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/progress/ProgressIterator.java
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @throws NoSuchElementException - if the iteration has no more elements
+     * @throws IllegalStateException is the Iterator and therefor the
+     *         ProgressBar have been closed.
+     * @return
+     */
+    @Override
+    public E next() throws NoSuchElementException {
+        bar.update();
+        return inner.next();
```
dkfellows: I'd prefer this as it updates the bar _after_ the value has been computed:
```
try {
    return inner.next();
} finally {
    bar.update();
}
```

SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/UnitConstants.java
Christian-B: True. Could add days here when and if needed.

SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/progress/ProgressIterator.java
Christian-B: Nothing wrong with that but the difference will be extremely minimal in nearly all cases where the value just comes from a collection.

I wonder if what we really want is to update after the caller is finished doing whatever it does with the E.

SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/UnitConstants.java
Christian-B: Yep test confirm hours > 24 gives WRONG result.

Simpler solution used which gives any number of hours.

SpiNNaker-utils/src/main/java/uk/ac/manchester/spinnaker/utils/progress/ProgressIterator.java
Christian-B: Did a new solution which waits with the progressbar update until the caller is finished with the returning E.
this is detected by either.  
Another next() call
A hasNext() which results in a false, and therefor calling close.
A close() call   (for example by breaking out of a try with resources)  
         See TestIterable.testStopEarly()

PR #20, opened 2018-08-22
Added dependency mappings

Christian-B:
Adds Pom dependecies

- [ n/a] javadoc correct.
- [ n/a] test case(s).


dkfellows:
It's not exactly how I'd do it, but it's easiest to merge and then alter later when it matters.

PR #27, opened 2018-08-31
Dse

dkfellows:
Implementation of the **Data Specification Executor**. The DS Generator is deliberately _not_ done (other than the extremely primitive/cut-down version for testing); it needs to be coupled closely to the Python code that drives it. 

- [x] javadoc correct.
- [x] test case(s), ported from original ones used to test the Python code.


dkfellows:
This would be cleaner except that I discovered when writing it that a `ByteBuffer` doesn't carry its endianness over when you get a read-only view or duplicate of it. Because that would be too good, I guess…

SpiNNaker-data-specification/src/main/java/uk/ac/manchester/spinnaker/data_spec/Constants.java
```diff
@@ -0,0 +1,85 @@
...
+	private Constants() {
+	}
+
+	/** Bytes per int/word. */
+	static final int INT_SIZE = 4;
+	/** Bytes per long/double-word. */
+	static final int LONG_SIZE = 8;
+
+	/** Data spec magic number. */
+	static final int DSG_MAGIC_NUM = 0x5B7CA17E;
```
Christian-B: I suggest linking to DSG documentation that uses the name "Magic num"

SpiNNaker-data-specification/src/main/java/uk/ac/manchester/spinnaker/data_spec/Executor.java
```diff
@@ -0,0 +1,197 @@
...
+
+import uk.ac.manchester.spinnaker.data_spec.exceptions.DataSpecificationException;
+
+/**
+ * Used to execute a SpiNNaker data specification language file to produce a
+ * memory image.
+ *
+ * @author Donal Fellows
+ */
+public class Executor implements AutoCloseable {
```
Christian-B: I suggest implementing Closeable  

The close method only throw IEException and is idempotent.

PR #30, opened 2018-09-05
Fixes for endianness of byte buffers.

dkfellows:
Because duplicated and as-read-only buffers _must_ start as big endian even if the buffers they're coming from are little endian. Of course!


dkfellows:
No longer relevant

PR #36, opened 2018-10-02
Fix broken SCP serialization

dkfellows:
Needs two extra leading zero bytes to match [`scamp_connection.py`, line 67](https://github.com/SpiNNakerManchester/SpiNNMan/blob/4d8c85fb878dc7a636ccb23cfc9f00ede90d7303/spinnman/connections/udp_packet_connections/scamp_connection.py#L67).


Christian-B:
That looks better!

PR #40, opened 2018-10-08
DB storage

dkfellows:
This is a storage system based on an SQLite database. It supports both in-file and in-memory versions (the latter won't be useful for transferring to another process…)

Key files:

* `/SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/DatabaseEngine.java` — how to connect to the DB
* `/SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/SQLiteStorage.java` — how to use the DB once you have a connection (includes DML and DQL)
* `/SpiNNaker-storage/src/main/resources/db.sql` — the definition of the DB's tables (DDL)


dkfellows:
This is now good to go if we accept that what it's doing is correct in the first place.

Adding in some sort of caching of connections (as a performance measure) would be a good idea, but I'll do that as a separate PR.

PR #45, opened 2018-10-10
Throttle p2pColumn

Christian-B:
fixes incorrect column data


Christian-B:
For some reason I can not even start to try to explain this runs faster on my ubuntu machine inside of netbeans.

Error has gone from about 1 in 5 times to not detected in 1500 tries on my machine and rare in travis so merging!

PR #48, opened 2018-10-12
Stop reordering

### LINKED ISSUE #43: More reliable getmachineinfo
Sometimes a read from a chip fails. (Noticed during getting the description of the machine.) That should result in at least one attempt to re-read it.

dkfellows:
Fixes #43 which is equivalent to SpiNNakerManchester/SpiNNMan#141, and which shows up in Java much more readily because Java is much faster than Python.


Christian-B:
Tested this and the fix worked.

However setting the THROTTLED to 8 greatly slows down the ReliabilityITCase from running in 3 seconds to running for 1.4 minutes.

So have changed it back to 3 after doing full testing at 8.

May be a case where a config reading is actually the most desirable.

PR #57, opened 2018-10-30
Byte debug

Christian-B:
better printing of ByteBuffers.

hope we NEVER need this again but just in case.


Christian-B:
Woops way way to much stuff here

PR #60, opened 2018-11-06
Fix two faults in large memory reads

dkfellows:
1. The maximum size of memory transfer unit is 255 (it goes through a `uint8_t` or `byte` somewhere)

2. The maximum size of data to transfer must not exceed the number of bytes remaining to be transferred.


Christian-B:
ok to merge if green!

PR #62, opened 2018-11-07
Improve consistency of request pipelines

dkfellows:
Sequence numbers can wrap. That causes trouble when packet sending is fast enough. Fixing that also lets us reduce the number of internal state variables, which is itself a *GREAT* thing.


Christian-B:
Rubbered when Green!

PR #65, opened 2018-11-13
Allow memory reads to funnel straight to the database

dkfellows:
This is about allowing reads from memory to be funneled more efficiently into the database.

It's tricky because we can't (easily) write to the middle of a BLOB with the DB API we have, so we instead need to stream things in when we can. And take care with out of order messages.


dkfellows:
This is still a Work In Progress

Christian-B:
Like this idea.

May be worth looking at DataReceiver.getDataForPlacementLocked and how this fits in.

PR #94, opened 2018-11-30
Indentation in checkstyle

Christian-B:
fixes so strange indentations

Requires some code cleanup to go green.

PR #96, opened 2018-11-30
Indentation2

Christian-B:
builds on https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/94
fixes all but Communications


dkfellows:
Handled via #94

PR #100, opened 2018-12-03
Make the Transceiver's machine field private

### LINKED ISSUE #99: Transceiver.machine
_Remember: the better you describe your issue, the easier it is for us to help you._

What is the problem you see?
dont like it being package protected, should be private
============================

What do you see that is wrong?
 package protected, should be private

access only at that level for test. should utilise back door

master
-------------------------------------------

dkfellows:
Fixes #99


alan-stokes:
waiting on @Christian-B to be happy

Christian-B:
Waiting on travis-ci.org to be happy!

Issue #87, opened 2018-11-29
magic number and odd result


alan-stokes:
_Remember: the better you describe your issue, the easier it is for us to help you._

What is the problem you see?
RoutingEntry has a magic number in figuring NUM_PROCESSORS (aka 32) and to add, that calculation returns a result of 26, which at last check, was not how many cores a spinnaker 1 has
============================

What do you see that is wrong?
magic num and wrong resulting value

What do you think it should do instead?
should be 18 or MAX_NUM_CORES from MachineDefaults.java

What version of our software are you using?
master
-------------------------------------------

alan-stokes:
i dont think this has actually caused any issues, as the cores from machine shouldnt never go above 18. but its a opening for dubious.

alan-stokes:
but 8 bits of that are not used by cores. not sure what those 8 bits are used for. but we obvs dont have cores 19 to 26

dkfellows:
Had to hunt around: in `spinnaker_tools/sark/sark_hw.c` in line 302, you can see that the top eight bits are used for the `app_id`. Without masking.

We'd better make sure that those bits are zero when setting them in here.

Christian-B:
@dkfellows is this fixed now?

dkfellows:
Looks fixed to me; I can't see any way to use out-of-range processor IDs now.

PR #110, opened 2018-12-03
Use machine defaults constants to clamp range of route bits

dkfellows:
Fixes issue #87

The problem is that sark (which SCAMP is acting as thin interface to) doesn't mask the bits out when it sets them, and instead just adds in the `app_id` into them; client code _must not_ set the top 8 bits. Given that we need to clamp properly, might as well use the constants defined elsewhere.

Need to fix in SpiNNMachine too.


Christian-B:
looks ok when green

Issue #194, opened 2018-12-07
mutible


alan-stokes:
Full path to where you find the problem:
C:\Users\alan\Documents\software\JavaSpiNNaker\SpiNNaker-machine\src\main\java\uk\ac\manchester\spinnaker\machine\RoutingEntry.java

What is the problem you see?
    /**
     * Removes a link ID/Direction if it existed.
     *
     * @param oldValue The Id of the processor to remove.
     * @return {@code true} If this entry contained the specified
     *      link/direction.
     */
    public boolean removeLinkID(Direction oldValue) {
        return linkIDs.remove(oldValue);
    }

What do you think it should do instead?
this makes the class mutable. Which is against our mantra about spinnMachine. Its only ever used by 1 test, which makes it even more stinky. 

remove the test and the method

What version of our software are you using?
master

Christian-B:
This issue is incomplete!

Removing only one of two remove methods and leaving the public add methods does not change the Object Immutable.

Christian-B:
Mutibilty was a left over from when the Fiddler removed bad links.

PR #199, opened 2018-12-07
removed mutibility.

Christian-B:
fixes https://github.com/SpiNNakerManchester/JavaSpiNNaker/issues/194

### DIFF
--- a/SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/RoutingEntry.java
+++ b/SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/RoutingEntry.java
@@ -59,8 +59,19 @@ public class RoutingEntry {
      */
     public RoutingEntry(
             Iterable<Integer> processorIDs, Iterable<Direction> linkIDs) {
-        addProcessorIDs(processorIDs);
-        addLinkIDs(linkIDs);
+        for (int procId : processorIDs) {
+            if (procId >= MAX_NUM_CORES || procId < 0) {
+                throw new IllegalArgumentException(
+                    "Processor IDs must be between 0 and "
+                            + (MAX_NUM_CORES - 1) + " found " + procId);
+            }
+            if (!this.processorIDs.contains(procId)) {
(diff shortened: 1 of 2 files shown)


SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/RoutingEntry.java
```diff
@@ -59,8 +59,19 @@ public RoutingEntry(int route) {
...
+            if (procId >= MAX_NUM_CORES || procId < 0) {
+                throw new IllegalArgumentException(
+                    "Processor IDs must be between 0 and "
+                            + (MAX_NUM_CORES - 1) + " found " + procId);
+            }
+            if (!this.processorIDs.contains(procId)) {
+                this.processorIDs.add(procId);
+            }
+        }
+        for (Direction linkIds: linkIDs) {
```
dkfellows: Would it be better to make the collection of directions (and possibly of processors) into something other than a plain `ArrayList`? For example, a `LinkedHashSet` or a `TreeSet`?

alan-stokes: the only thing i can see which might affect that decision, is the two apis for getProcdessorIDs and getLinkIDs. These both expect a List<Direction> and List<Integer>. So converting between a ArrayList and List is a bit trivial. LinkedHashSet or TreeSet...... maybe not so trivial. 

inside the class, its only ever adding, or iterating over the lists. so maybe its not worth changing?

alan-stokes:
so from my point of view, this is fine to merge. leaving for @dkfellows  to decide if hes happy

SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/RoutingEntry.java
dkfellows: Hmm. Thinking about this, the stuff I'm thinking about should be a separate change.

PR #207, opened 2018-12-10
Style changes

### LINKED ISSUE #101: comments and code and new lines
_Remember: the better you describe your issue, the easier it is for us to help you._

This is more a place holder for us to dicuss. not really a wrong or right. 
there are currently 2 conventions happening in the java code.

    /** comment*/
    code
    /** comment*/
    code

and

    /** comment*/
    code

    /** comment*/
    code

(note the new line between new blocks). Persoanlly i find the second more readable than the first, and most of the java ive seen to date does the second, but the comms seems to follow the first. 

discuss if we want a unified convention here, or meh to either. Im not sure, so leaving it as issue for opening debate

dkfellows:
fixes #101 

Style changes are _large_.


alan-stokes:
failed checks. will review when ya ready (aka when ticks are green)

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/eieio/EIEIODataMessage.java
```diff
@@ -380,11 +388,17 @@ public int getSize() {
 		}
 
 		private static final int PREFIX_BIT = 7;
```
alan-stokes: These should be at top of file.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/model/RouterDiagnostics.java
```diff
@@ -41,23 +41,32 @@
 public class RouterDiagnostics {
```
alan-stokes: i think (but am really only slightly off the fence on this one, that the enum should be at top of file.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/model/Signal.java
```diff
@@ -54,8 +54,10 @@
```
alan-stokes: for the ones above this. should there not be new lines between the varable and next comment???

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/model/SystemVariableDefinition.java
```diff
@@ -209,11 +209,13 @@
```
alan-stokes: same for the ones above here

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/scp/IPTagGet.java
```diff
@@ -104,11 +112,17 @@ private Response(ByteBuffer buffer)
 		}
 
 		private static final int IPV4_BYTES = 4;
```
alan-stokes: mov above the private response?

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/transceiver/UDPTransceiver.java
```diff
@@ -152,6 +155,7 @@ public void close() throws Exception {
 	}
 
 	private static final InetAddress WILDCARD_ADDRESS;
```
alan-stokes: this still has the issue of multiple methods between the 2 uses, so psuh to top of class

Christian-B:
The standard which I agree with is that PRIVATE variables even static ones do not need javadocs as long as by their name it is relatively clear what they represent.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/model/RouterDiagnostics.java
dkfellows: Java _mostly_ doesn't care, and reordering class contents is off-topic for this PR.

dkfellows:
This will need to be redone if we want it; `master` has moved too far.

PR #208, opened 2018-12-11
Better underlying types

dkfellows:
This branch was inspired by looking at our uses of the `contains()` operations of collections; virtually everywhere where that crops up, it turns out that we should be using sets rather than general collections or lists. _Technically_ it's quite an intrusive change as it is visible in a lot of APIs, but practically I don't think it is such a big deal.

After all, we want to encourage such lookups as happen to be approximately O(1) rather than O(N), especially when building the model of the real machine…


rowleya:
Just a minor note - a Set is a Collection, so you don't have to change the interface unless you need something specifically in the Set interface (contains() is in Collection).  This usually means you just need to change the part that assigns the collection to use HashSet instead of ArrayList (or whatever set implementation you want).  Using Collection is normally better if there isn't any other contract...

dkfellows:
We discussed this; I prefer to use a set when `contains` method is the main point of the object and I tend to use interfaces even for internal variable types.

PR #209, opened 2018-12-11
DSE exec

dkfellows:
This is the parallel DSE.

Technical points:
* Base branch for this is `db-change-2`, not `master`. ***BEWARE!***
* Chosen package name: `uk.ac.manchester.spinnaker.front_end.dse`. I'm not very attached to it.
* The parallelization factor (number of boards accessed simultaneously) is 4 (guesswork!), and is set in `uk.ac.manchester.spinnaker.front_end.dse.HostExecuteDataSpecification.PARALLEL_FACTOR`.
* I've had to extend the DSE database schema to add an `app_id` column to the `core` table; the application identifier is required _by SCAMP_ to do memory allocation.


dkfellows:
I'm also not sure what arguments ought to be passed over when creating a transceiver.

rowleya:
If you have already booted boards, the only thing a transceiver should need is the address of the board.

Note that the parallel factor should probably be based on the number of boards.  A simple way to do this would be to have the arguments to this be a list of IP addresses, and then the parallel factor would be the size of the list.

dkfellows:
Good. In that case I picked the correct constructor and passed the right arguments. 😁

SpiNNaker-storage/src/main/resources/dse.sql
```diff
@@ -33,6 +33,7 @@ CREATE TABLE IF NOT EXISTS core(
 	y INTEGER NOT NULL,
 	processor INTEGER NOT NULL,
 	board_id INTEGER NOT NULL,
+	app_id INTEGER NOT NULL,
```
Christian-B: Ah app_id  !

alan-stokes: yea i forgot about that one old boy. soz

PR #210, opened 2018-12-13
Database change

dkfellows:
This is the database change that matches the current plans for the buffer manager. Builds on top of the `extractor2` branch.

***Depends on #209***, which is the data specification executor to go with this, though this can be reviewed first.


SpiNNaker-storage/src/main/resources/buffer_manager.sql
```diff
@@ -0,0 +1,46 @@
...
+-- A table describing recording regions.
+CREATE TABLE IF NOT EXISTS region(
+	region_id INTEGER PRIMARY KEY AUTOINCREMENT,
+	core_id INTEGER NOT NULL,
+	local_region_index INTEGER NOT NULL,
+	address INTEGER NOT NULL,
+	content BLOB NOT NULL DEFAULT X'',
+	fetches INTEGER NOT NULL DEFAULT 0,
+	append_time INTEGER,
+	FOREIGN KEY(core_id) REFERENCES core(core_id) ON DELETE CASCADE);
```
Christian-B: Why not do ON DELETE RESTRICT   
I know we never delete so that is safer.

SpiNNaker-storage/src/main/resources/dse.sql
```diff
@@ -0,0 +1,49 @@
...
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+-- A table describing the boards.
+CREATE TABLE IF NOT EXISTS board(
+	board_id INTEGER PRIMARY KEY AUTOINCREMENT,
+	ethernet_x INTEGER NOT NULL,
+	ethernet_y INTEGER NOT NULL,
+	address TEXT UNIQUE); -- NB: NULL *or* UNIQUE
```
Christian-B: I understand that there are cases where it is possible with address null 

However in this case I would not allow it.

Instead have the system list the chips on various physical boards to the ethernet that does the talking.

SpiNNaker-storage/src/main/resources/dse.sql
```diff
@@ -0,0 +1,49 @@
...
+
+-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+-- A table describing the cores and the DSE info to write to them.
+CREATE TABLE IF NOT EXISTS core(
+    core_id INTEGER PRIMARY KEY AUTOINCREMENT,
+	x INTEGER NOT NULL,
+	y INTEGER NOT NULL,
+	processor INTEGER NOT NULL,
+	board_id INTEGER NOT NULL,
+	content BLOB,
```
Christian-B: app_id?

dkfellows: In #209

SpiNNaker-storage/src/main/resources/buffer_manager.sql
dkfellows: 👍

SpiNNaker-storage/src/main/resources/dse.sql
dkfellows: Are we going to guarantee that's `NOT NULL`? I'm happy with that…

Christian-B: I base this on the use case of using the board/ ethernet chip to load the data specs.

You can not load over a null address so the code has to provide an ethernet.
If there are multiple physical boards sharing a single ethernet connection then for the purpose here they must be considered as one "board"

Christian-B: If a chip's nearest ethernet has no address I will use probable only Ethernet board with 0,0

If there are multiple ethernet boards (but not all) I will use 0,0 and log a warning

if there is one ethernet board but not at 0,0 I will use that.

if there are multiple ethernet boards but none at 0,0 the system goes pop. 






If

alan-stokes: this has issues, unless you go and update ther chips nearest x and nearest y for ethernets. currently noone updates it, and there may be use cases where thats used to forward packets.

Christian-B: There is currently not a plan to update the chip nearest ethernet x and y although maybe we should.

As this database is specifically designed for ds transmission for the physical boards without an ip address the database must replace the ""board" as described above.

PR #211, opened 2018-12-13
Buffermanager 

Christian-B:
Dead code kept for reference just in case.

IE not ready to nuke branch yet.


Christian-B:
probably never to be merged so closing PR but NOT deleting branch,

PR #221, opened 2019-02-19
Scamp crasher

Christian-B:
Locks up scamp by sending messages in both directions at the same time.

DO NOT MERGE


Christian-B:
MARKER LEAVE BRANCH BUT NEVER MERGE.

Use exctractor2 branch which has a Java side fix!

PR #224, opened 2019-04-09
fixes #223

### LINKED ISSUE #223: Synchronization ERROR!
Full path to where you find the problem:
uk.ac.manchester.spinnaker.front_end.download.storage_objects.BufferedReceivingData

What is the problem you see?
endBufferingSequenceNo getting lost

Synchronization
quick fix:
        endBufferingSequenceNo = Collections.synchronizedMap(new HashMap<>());
        endBufferingState = Collections.synchronizedMap(new HashMap<>());
better fix
        synchronize only the setters that could add a key

long fix:
       Check all threaded access to shared maps

master

dkfellows:
fixes #223 

Synchronized maps pay a somewhat higher price, and yet we won't get much benefit from them. Instead, since the keys aren't actually shared between threads, we can just synchronize anything that can create an entry. And avoid using two calls into the maps where one will do.


Christian-B:
merge when green!

But then nuke https://github.com/SpiNNakerManchester/JavaSpiNNaker/tree/jenkinsjava

PR #235, opened 2019-07-26
Combine processes

dkfellows:
1. Reduces the number of process classes and moves them into the transceiver package, since that was the only possible user of them, and reduce the scope of the moved classes so that they're out of harm's way.
2. Makes `ProcessException` a subclass of `SpinnmanException` since the latter is clearly intended to be a super-exception.
3. Adds in some operations (as yet unused by high-level code) to fill out the set of operations supported by the transceiver. (This includes defining how one-way messages work at the transceiver level; previously, they simply didn't work in horrid ways…)


dkfellows:
Combined into #248

PR #243, opened 2019-11-19
changes to the data in protocol to reflect the changes in the python and

alan-stokes:
c according to chats from rowley
reflects changes in FEC on both c and python

https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/512
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/308


dkfellows:
Branch not deleted: code cleanup in progress. Branch merged because working when `master` bust!

PR #245, opened 2019-12-10
Neuron recorder agdr

rowleya:
This changes a value from < 256 to <= 256, a critical change in the current master code.

This also adds a log message that was very useful to debug (since it aligns with the logs from the extra monitor core).


alan-stokes:
can you point me at whatever the real change is. as its hidden under the tabs movement. Just in case its inportant to my stuff

rowleya:
Ah yes, try this (hides white space changes):
https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/245/files?utf8=%E2%9C%93&diff=split&w=1

rowleya:
No longer relevant

PR #255, opened 2020-02-07
Tighter style for exceptions

dkfellows:
Following on from commit 1f5f549, this tightens up the style rules so we can use what aggressive exception documentation checking that checkstyle actually provides


dkfellows:
(See https://github.com/checkstyle/checkstyle/issues/7329 for why this branch is necessary. It's a pain…)

dkfellows:
This fixed #253

PR #257, opened 2020-02-10
Remember missing

rowleya:
Misnamed now - was going to remember the missing for retransmit, but realised that this is not a good idea!

Instead, this now keeps a state to know whether we ever got anything back from the machine.  We shouldn't resend everything if we have ever received something back as that is wasteful and not likely to help.  Note that after a retransmit, there is no need to send all the missing packets again as we should just trigger a "send missing" at this point.  Additionally, we don't need to send a "location" here again since at this point we now know that the other end has received the first location, as we have received at least some missing sequences by this point.

Note that this still suffers from SCAMP losing response packets, meaning that often a retransmit will result in a timeout and another request for missing packets.  This seems to be the most common case, similar to the issue with sliding windows.  Thus this needs to be fixed as well (to be done in spinnaker_tools of course).  The result of this is that the protocol often ends up sitting in a "waiting for timeout" state, meaning that it will seem slow particularly on small data transfers.


dkfellows:
Changes effectively already on `master`.

alan-stokes:
as a side argument. resending the missing seqs shouldn't be a problem. as its not there's going to a new missing seq from that set. so our just giving the extra mon more chances to get those missing seqs. 

why would scamp lose the response packet? given by the time the exta mon responds, there shouldn't be any other messages flying around?

also. this pr seems to have no gains in it..... par a log_debug.

alan-stokes:
(cwl) donal beat me to it

PR #258, opened 2020-02-12
Use the default UDP send buffer size

dkfellows:
It turns out that this is *larger* than the value we were using (at least on macOS — where it defaults to 9216 rather than the value we were using of 1500 — and probably on Linux and Windows too), and this is useful when the network is a bit congested.

----
Background:
--
There's a _lot_ of confusion in this area (especially as UDP is far less understood than TCP by general , but it is important to remember that the value for `SO_SNDBUF` is used by the OS and not by Java (or other user code) to determine the size of the buffer before the data hits the network hardware, and the buffering can be required either because the network itself is busy or because there's other activity going on on the machine, or because there needs to be an ARP handling for the address, or … and because we're in non-block mode, the only option for dealing with a full buffer is to drop the packet.

* https://stackoverflow.com/questions/28692511/so-sndbuf-and-so-rcvbuf-settings-for-udp
* https://stackoverflow.com/questions/16156326/so-sndbuf-and-so-rcvbuf-on-linux-sockets

I hate dealing with this level of ignorance, which often occurs when you're in the close vicinity of something popular but don't actually need the popular thing.


rowleya:
Didn't notice this was a PR - thought it was an issue!  Anyway, this has been done in a different PR now so closing.

PR #261, opened 2020-02-13
Fewer timeouts

rowleya:
These changes result in fewer timeouts when using the fast data in protocol.  This correspondingly improves the speed of operation.

This is achieved by:
 - Use the default send buffer size
 - Check that data is actually sent on send
 - Fix what happens on a timeout (don't send all the packets if you have heard from the machine in this transaction).


rowleya:
Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/352

PR #269, opened 2020-10-06
Use right exception type on no-such-host

### LINKED ISSUE #215: TestUDPConnection unexpected exception type
Full path to where you find the problem:
uk.ac.manchester.spinnaker.connections.TestUDPConnection (in tests of course)

What is the problem you see?
org.opentest4j.AssertionFailedError: Unexpected exception type thrown ==> expected: <java.net.SocketTimeoutException> but was: <java.net.SocketException>
	at uk.ac.manchester.spinnaker.connections.TestUDPConnection.testSendSCPRequestToNonexistentHost(TestUDPConnection.java:99)

More general exception

expander2

dkfellows:
Fixes #215

Also clean up a few places where my IDE was grumbling


Christian-B (approved):
I see no problems here

PR #270, opened 2020-11-09
Paper tests

alan-stokes:
this adds functionality for tracking values for the speed up paper. probabily needs a mega clean up


dkfellows:
This branch has not been sufficiently maintained to be likely to be simply mergeable in the foreseeable future.

PR #271, opened 2020-11-11
Define a build action

dkfellows:
Includes all the magic for coverage reporting. Which is ghastly complicated!


dkfellows:
See also https://github.com/trautonen/coveralls-maven-plugin/issues/136#issuecomment-665233584

PR #273, opened 2020-11-13
Generate and publish docs

dkfellows:
Enables generation of a “Maven site” for JavaSpiNNaker, and publication of that site to github.


rowleya (approved):
Sounds like a good idea to me!

PR #275, opened 2020-12-07
Add uploading of the main build artifact, and schedule a daily build

dkfellows:
This will let people get at builds of this code without needing to figure out Maven.


rowleya (approved):
Lets get it out there

dkfellows:
:shipit:

dkfellows:
(Don't ask me why `:shipit:` is :shipit:, but it is.)

PR #276, opened 2020-12-17
Split the long build+test from the quicker validate+doc

dkfellows:
Mark non-critical steps as able to fail silently (as they're just best-effort)


rowleya (approved):
Looks like a good idea to me!

PR #278, opened 2021-03-03
Try to prevent catastrophes from occurring

dkfellows:
I suspect some sort of weird thread funkiness. Let's be careful...


rowleya (approved):
Yes, full sync is a good idea!

rowleya:
Worth an integration test, just to be sure?

Christian-B:
> Worth an integration test, just to be sure?

Conisdering  branch dist_dep_weights_delays failed twice in the same way may be worth testing against that

rowleya:
A branch has been made of this with the name dist_dep_weights_delays to test...

dkfellows:
If you _can_ test this, please do. Threading-related issues are disgustingly difficult to test…

rowleya:
It was more of a "does it fail on some odd test case now it has changed"; if all tests pass on that branch I think that can be called success, even if it wasn't actually because of this change!

PR #279, opened 2021-03-04
Simplify recording

rowleya:
Changes that are needed by SpiNNakerManchester/SpiNNFrontEndCommon#768.


andrewgait (approved):
This looks fine to me but my Java is not up to the standard of @dkfellows or @Christian-B so I suggest one of them check it over too to make sure it's good.

dkfellows:
I want to tweak this a bit.

dkfellows (approved):

PR #282, opened 2021-03-25
Create a security policy

dkfellows:
This creates a security policy that will show up specially identified as such in the GitHub UI.


rowleya (approved):
Looks good

PR #285, opened 2021-03-25
Log problem boards

dkfellows:


dkfellows:
Nice idea, but unlikely to ever be useful because the update required to the (old) spalloc server isn't likely to happen.

PR #288, opened 2021-04-08
Create dependabot.yml

dkfellows:
This will enable keeping track of versions of things we depend on because otherwise we don't have visibility of this sort of thing.


dkfellows:
Note that this is *not* going to be part of the 6.0 release

PR #312, opened 2021-04-20
Bob the Builder

dkfellows:
Can we fix it?
Yes we can!

----

Trying to make the Java API to Spalloc less obnoxiously Pythonic; Java much prefers a fluent builder interface for this sort of thing. It's still a ghastly API, but now it at least makes the things it can do more discoverable and provides some basic interlocking. Not yet removed the old way of working…


Christian-B (approved):
Assuming this works looks good.

Ok to merge

PR #320, opened 2021-04-27
Style changes

dkfellows:
More features from checkstyle so that we can have a less strict style policy for test code (e.g., to allow magic numbers). But we do add in one key piece of strictness (actually several, but the others are no big deal): fields (NB: not local variables) must have a blank line between them. To see what I mean, look at this:
```java
class Foo {
    // bad
    int x;
    int y;
}
```
```java
class Foo {
    // also bad
    int x, y;
}
```
```java
class Foo {
    // good
    int x;

    int y;
}
```

And added a bunch more tests.

----

Most of the commits on this branch are due to fixing the style due to the above change.

PR #321, opened 2021-04-28
More spalloc tests and API improvements

dkfellows:
Extracted from #320 where they're a bit off-topic; do this PR first.


Christian-B (approved):
Approved

PR #324, opened 2021-05-07
Create .editorconfig

dkfellows:
Apparently this lets us set the right tabs and stuff like that on a per-project basis

https://github.com/isaacs/github/issues/170#issuecomment-150489692

> When you have a `.editorconfig` in your repository it will respect it when viewing code on GitHub.
>
> `indent_style = tab` and `indent_size = 4` shows tabs with 4 columns instead of 8

[This](https://gist.github.com/JamesMGreene/f4b01ff4a2f8e6faead1) was also useful.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,27 @@
+# Copyright (c) 2021 The University of Manchester
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


dkfellows:
Well that seems to work, at least when viewing the POM file.

PR #325, opened 2021-05-07
Spalloc server

dkfellows:
This is a completely new implementation of Spalloc Server. It uses a new protocol (conventional REST) and supports a web interface for most client uses (still a bit of a work in progress). It has _optional_ support for the old protocol, but I'd like that to end up turned off.

It's all database-backed, so service restarts should be painless and crashes ought to not lose anything that's been committed to.

The new protocol requires authentication. Only the utter horribleness of the old one allowed us to get away without doing this before.


dkfellows:
TODO: Look at using `RETURNING` now that we have a new enough SQLite.

dkfellows:
OK, I think that's a full local security solution that is properly careful with things like password exposure.

dkfellows:
Love it when test dependencies change their internal structure such that problems only show up on this branch. 😞

dkfellows:
Configuration Notes
====
Note well that I've omitted some sections from the `application.yml` file. Those should be left alone by default!
```yml
spring:
  mvc:
    servlet:
      # Part of the configuring of what paths are where within the service. This is
      # changeable, but may require other bits and pieces to be altered as well
      path: /system
  task:
    scheduling:
      pool:
        # Default thread pool size. Note that the thread pool is shared by many
        # things inside the service; if this value is too small, performance
        # may be severely impacted or requests may fail.
        size: 16
  mail:
    # Configuring email is fiddly; see docs for how. WE WANT TO CHANGE THIS FOR UNIMAN
    # See https://www.baeldung.com/spring-email
    # See https://docs.spring.io/spring-boot/docs/2.0.6.RELEASE/reference/html/boot-features-email.html
    host: smtp.gmail.com
    port: 587
    username: "${EMAIL_USER:}"
    password: "${EMAIL_PASS:}"
    properties:
      "[mail.smtp.auth]": true
      "[mail.smtp.starttls.enable]": true

server:
  # What local port to run on; NOT the address of any nginx front-end
  port: 8080

cxf:
  # Part of the configuring of what paths are where within the service. This is
  # changeable, but may require other bits and pieces to be altered as well
  path: /srv
  servlet.init: 
    # Part of the configuring of what paths are where within the service. This is
    # changeable, but may require other bits and pieces to be altered as well
    service-list-path: info

# Don't set pause, auth.add-dummy-user, transceiver.dummy or auth.debug-failures in production
spalloc:
  # Where's the database?
  database-path: spalloc.sqlite3
  # How long should long calls take before returning anyway?
  wait: 30s
  # Whether to pause ALL periodic callbacks; probably only useful for debugging
  pause: false

  historical-data:
    # Where to we back up old jobs to?
    path: spalloc-history.sqlite3
    # How long before we move old jobs to the historical data
    # Defaults to 14 days (or thereabouts)
    grace-period: 14d
    # Cron-like expression to say when to run the migration task.
    # Defaults to overnight at 02:00 every day
    schedule: "0 0 2 * * *"

  keepalive:
    # How long between runs of the keepalive-expiry algorithm
    expiry-period: 30s
    # Minimum keepalive period
    min: 30s
    # Maximum keepalive period
    max: 300s

  allocator:
    # How long between runs of the allocator algorithm
    period: 5s
    # What span of importance will be considered when allocating jobs
    importance-span: 10000
    # Priority is the rate at which importance is accrued
    priority-scale:
      # Priority scaling factor for jobs specified by number of boards
      # Will be multiplied by the number of boards
      size: 1.0
      # Priority scaling factor for jobs specified by rectangular dimensions
      # Will be multiplied by the area of the allocation
      dimensions: 1.5
      # Priority scaling factor for jobs requesting a specific board
      specific-board: 65.0
    # Number of reports of board problems at which the board is taken out of service.
    report-action-threshold: 2
    report-email:
      # In addition to configuring the email sender, you need to turn on
      # sending here too.
      send: false
      from: spalloc@localhost
      to: SPINNAKER@listserv.manchester.ac.uk
      subject: Spalloc Board Issue

  auth:
    # Support HTTP Basic Authentication (bad for browsers, OK for code)
    basic: true
    # Support HTTP Form Authentication (pretty necessary)
    local-form: true
    # Force a known local admin user to exist with a known password
    add-dummy-user: false
    # Whether to generate a random password for the above user.
    # If so, the password will be written to the log.
    dummy-random-pass: false
    # Provide extra information to callers on auth failures
    debug-failures: false
    # Number of login failures before lock-out
    max-login-failures: 3
    # Duration of automatic log-out
    account-lock-duration: 24h
    # How often do we look for users to unlock
    unlock-period: 60s

  quota:
    # Default user quota in board-seconds
    default-quota: 130
    # When do we consolidate job quotas into the main quota table
    consolidation-schedule: "0 0 * * * *"

  transceiver:
    # How long between when we send requests to the BMP control tasks
    period: 10s
    # The basic wait time used by the BMP control tasks
    probe-interval: 15s
    # Number of attempts that will be made to switch on a board
    power-attempts: 2
    # Number of attempts that will be made to bring up an FPGA
    fpga-attempts: 3
    # Whether to use dummy BMPs; GREAT for testing!
    dummy: false

  sqlite:
    # How long to wait to get a DB lock
    timeout: 1s
    # Whether to send details of SQL-related exceptions to users
    debug-failures: false
    # Amount of effort to spend on DB optimization on application close
    analysis-limit: 400

  # Settings relating to the v1 spalloc compatibility interface
  compat:
    # Whether to turn the service interface on
    enable: false
    # The number of threads to use for the API (including notification handlers)
    # Use 0 for no hard limit.
    thread-pool-size: 0
    # What port to run the service on
    port: 22244
    # What user to run as; recommended that this user exists but is disabled
    service-user: ""
```
There will be additional settings when the OIDC support branch is merged (critically including the bits for authenticating the service to the EBRAINS auth service).

dkfellows:
See also https://github.com/SpiNNakerManchester/spalloc/pull/62

dkfellows:
Here's a screenshot I took a few minutes ago, with the original request (which default to being hidden) and both timestamps and intervals pretty printed. The mouse pointer is over the yellow hexagon, so a tooltip is showing for it (indicating the coordinates — usually not too interesting — and the IP address of the allocated board). The job is owned by `spalloc-classic`, a user I've made to hold all jobs submitted via the legacy interface.

screenshot

Note that `max_dead_links` and `require_torus` from the request are _always_ ignored, as they are never set via our tools anyway.

The corresponding machine view is this:

image

In the machine view, information about jobs is only shown via the popup over the map; clicking on the boards of the job takes you (provided you have permission to see the job's details) to the job view for that job. Jobs are colour coded “randomly” (via hashing of the job ID so that they have consistent if unpredictable colours).

I believe there will be some scaling problems with this view with the big machine; at the very least, I'll have to add code to make the map larger on the screen!

rowleya:
For the big machine, so long as the page would likely "grow" with scroll bars rather than trying to wrap, I don't think it will be too much of a problem.  It is already hard to visualise this on a terminal, so I can imagine that in a browser will almost certainly be better!

dkfellows:
The big machine is nearly impossible to visualise on a terminal, unless you use a small font and have the terminal full screen. Part of the problem is the wrong shape of hexagons being used (which is a problem I've corrected) and the other part of the problem is that the big machine is, well, big.

dkfellows:
That should be all the known `FIXME`s in the code addressed.

dkfellows:
Selection of screen grabs from the admin pages (shrunk here for ease of handling in this page; the titles are actually all the same size):

Screenshot 2022-02-08 at 12 36 27 Screenshot 2022-02-08 at 12 36 50 Screenshot 2022-02-08 at 12 37 27

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/security/LocalAuthProviderImpl.java
```diff
@@ -697,6 +697,14 @@ public void mapAuthorities(OidcUserAuthority user,
...
+		if (!collabToAuthority("userInfo",
+				user.getUserInfo().getClaimAsStringList("group"), results)
+				// Note: not a shortcut AND; always call both sides
+				& !collabToAuthority("idToken",
+						user.getIdToken().getClaimAsStringList("group"),
+						results)) {
+			log.info("no group in authority");
+		}
 		if (!orgToAuthority("userInfo",
 				user.getUserInfo().getClaimAsStringList("unit"), results)
```
rowleya, line 709: My question is whether this is "unit" or not.  Should it have been "team" and "group" rather than "team" and "unit"?

dkfellows:
Various admin views (as most will never see this):

# User Administration
Screenshot 2022-03-09 at 15 15 20
Screenshot 2022-03-09 at 15 15 48
Screenshot 2022-03-09 at 15 16 09

# Group Administration
Screenshot 2022-03-09 at 15 16 54
Screenshot 2022-03-09 at 15 17 20
Screenshot 2022-03-09 at 15 17 46

dkfellows:
A number of things in the above views are greyed out.

* Membership of OpenID groups (organisations and collabratories) is managed externally — synched for a user whenever they log in — so admins can't change those.
* OpenID users cannot have passwords set (or their subject ID). 
* Quota-less groups don't have the UI option to add or remove quota.
* Admins can't set their own password via the admin interface; they should use the same interface as ordinary users (as that takes care to verify existing credentials).
* Admins can't set many other aspects of their own status.

Things not shown:
* Both individual boards and whole machines can be taken out of service and put back in. This affects the UI and whether the hardware in question is available for allocation, but does *not* stop current use of it.

The UI *should* be scaled so that it works reasonably on a mobile device. I've tested with Chrome on a Galaxy running Android; YMMV. 😉

dkfellows:
All that really remains to do at this point is testing whether things actually work with the tools.

dkfellows:
I think this is up to merge quality now. Problems from here on can be their own bugs/PRs.

PR #331, opened 2021-05-25
Split synapse neuron models

rowleya:
The Java equivalent of SpiNNakerManchester/DataSpecification#108 and the DSE part of SpiNNakerManchester/SpiNNFrontEndCommon#815.


dkfellows (approved):
If it works, I'm happy with it.

PR #337, opened 2021-06-18
Publish maven packages 

dkfellows:
It'd be nice to publish the Java packages automatically. Unfortunately, publishing to Maven Central requires some DNS trickery to get started and I'm _really_ not willing to try to prod that side of things! But… Github hosts a Maven repository system too. That'd surely do for our basic needs.

This PR adds a workflow so that when we do a release (i.e., approximately whenever we tag a release) that release _automatically_ gets published.

----

See https://docs.github.com/en/actions/guides/publishing-java-packages-with-maven for details of what's going on in this PR.

### DIFF
--- a/.github/workflows/publish-pkgs.yml
+++ b/.github/workflows/publish-pkgs.yml
@@ -1,0 +1,20 @@
+name: Publish package to GitHub Packages
+on:
+  release:
+    types: [created]
+jobs:
+  publish:
+    runs-on: ubuntu-latest 
+    permissions: 
+      contents: read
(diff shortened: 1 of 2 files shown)


rowleya (approved):
Looks good.  I guess no one has anyone's credit card details so the worst that would happen if we go over quotas is that we would get a warning and this would stop working...

rowleya:
Side thought; if we ever get to the point that things are getting too big, there is also an action to delete packages:
https://github.com/actions/delete-package-versions

PR #347, opened 2021-06-30
Match data-spec changes for increase of max from 16 to 32

andrewgait:
Match changes in https://github.com/SpiNNakerManchester/DataSpecification/pull/112

Part of https://github.com/SpiNNakerManchester/sPyNNaker/pull/1076


rowleya (approved):
Yes, this looks like it should do the job!

dkfellows:
I find it curious that there was a spare bit, or maybe that's just something that's there in the commands that are implemented but not there in others. 😄

rowleya:
> I find it curious that there was a spare bit, or maybe that's just something that's there in the commands that are implemented but not there in others. 😄

I think that is the case;  because the command itself is identified separately, the bits used by each command are independent.

PR #351, opened 2021-07-08
Spalloc server JSP interface

### LINKED ISSUE #349: Web admin interface
There needs to be a web interface for doing administrative tasks. It _must_ be restricted to only logged in users with the `ADMIN` permission.

Current admin tasks that need UI:
1. Managing users and quotas
2. Uploading new machine definitions
3. Enabling and disabling a board in machine (technically just affects whether the board is available for allocation)

dkfellows:
This creates an interface for the server based on Spring MVC and JSP. I don't plan to use this for the main user interface as that needs to be much more dynamic (that'll have to be some sort of Javascript SPA), but administration can be done relatively statically.

fixes #349


dkfellows:
Still to do: forms for login and user self-maintenance. Those are non-admin forms; they'll need a different controller (but will be pretty standard MVC stuff).

Anyone who knows how to slap a nice style on this is _utterly_ welcome to do so. That's not a blocker on the PR though.

dkfellows:
Login form now done, and there's a (very!) basic stylesheet as well so we can handle any static resources we need.

Still need self-maintenance form.

dkfellows:
That all now seems pretty solid, and will do nicely as a base for the non-SPA parts of the user interface.

PR #358, opened 2021-08-20
Backport throttle update

rowleya:
Update the throttle value to work better.  The original value was working well for a while, but seems to now not work as well!  This new value appears to work better...


dkfellows (approved):
It probably depends on the speed of the machine and how the system timer actually works. In a VM running on updated hardware? 🤷

rowleya:
Oddly it even seemed to change on my own laptop connecting to a local board, where I didn't think anything else had changed at all.  In any case I don't think it breaks anything...

PR #380, opened 2021-11-25
Spalloc server: behind nginx

dkfellows:
Changes required so that the service sits behind its reverse proxy correctly. (It's really all about making sure that URLs are correct.)


dkfellows:
The detailed logging stuff was useful for a short while, but might have never been necessary if I'd known how to switch on the logging mode that surely already exists in the Spring Boot configuration. It was easier to write my own logger than to look up how to turn on the level of logging I needed…

rowleya:
Ah so many memories of spring logging issues...

dkfellows:
This is now a main part of the `spalloc-server` branch.

PR #385, opened 2021-12-08
Spalloc server: OIDC login

### LINKED ISSUE #338: Identify users
For #325, how should we identify users? How should we register users?

Old spalloc has no security at all. Everything is by self-asserted identity. Fortunately we've not had problems with bad actors.

### LINKED ISSUE #342: Security model
If we address #338, how will this affect the information returned?

My feeling is that owner, creation request and keepalive host information should only be returned to admins (about jobs that they don't own). Only owners and admins should be able to change a job (including keeping it alive) but general users _including unauthenticated users_ should otherwise be able to read the machines and jobs.

----

The restricted values should be restricted because of PII issues.

dkfellows:
fixes #338
fixes #342


dkfellows:
Can now log in with HBP/EBRAINS.

PR #401, opened 2022-01-11
Keycloak client

dkfellows:
The client for talking to Keycloak for doing the service installation setup.

I'm not sure if I'll ever merge this, but it is nice to not lose the code. Not really usable by anyone other than me, and very very partial in the functionality it supports.


dkfellows:
Note that the code in this module _requires_ Java 11 to build; it's profiled in the parent POM so that older Java versions won't even try.

PR #402, opened 2022-01-12
Better auth challenge

dkfellows:
Improve the auth challenge so that it says that both HTTP Basic and OIDC Bearer tokens are supported as auth methods.


dkfellows:
Need to check that this works in deployment before merging!

dkfellows:
I have literally no idea why Maven was choking on having the filter code as an inner class like that, but moving it to a new file was an ever-so-easy workaround.

PR #403, opened 2022-01-14
Make package so we can move the plethora of security bits out

dkfellows:
It's better this way, much more understandable. Only `SecurityConfig` itself is not moved (or bits that are extremely closely related to it), as there are critical order dependencies between things (in relation to the keystore installation and the startup of the OIDC support) and if things are working right now _then I'm not touching them_.


dkfellows:
The old way things have been arranged has been bothering me for a while.

dkfellows:
Need to test this before merging; initialization order for `SecurityConfig` (relative to the OIDC stuff) is critical.

dkfellows:
Works

PR #418, opened 2022-02-09
Spalloc server groups

### LINKED ISSUE #413: Group / Collab quotas
The HBP model of quota allocation assigns quota to collabs rather than individuals.  To make synchronization of quotas between the HBP service and the SpiNNaker allocation easier to manage, it would be helpful if the SpiNNaker allocation service could allocate quota to groups.  

Thoughts:
 - A user would need to be identified to be authorized to use quota from one or more groups.
 - A user can still have a default personal quota (which might be 0).
 - As part of a job, a user can identify which quota a job should be allocated from.  If the user is not authorized to use the quota, a quota of 0 can be assumed.  If the user doesn't identify a quota, their personal quota can be used by default.

dkfellows:
Implementation of groups for spalloc server (which will fix #413). Groups are the units to which accounting is performed, especially within the context of HBP.


dkfellows:
Against the wrong branch! 

image

PR #419, opened 2022-02-09
Spalloc server groups

### LINKED ISSUE #413: Group / Collab quotas

dkfellows:
Implementation of groups for spalloc server (which will fix #413). Groups are the units to which accounting is performed, especially within the context of HBP.

- [x] Core model changes
- [x] Database migration (or rebuild from scratch)
- [x] Link jobs so their accounting goes against their group
- [x] Create default group for local users
- [x] Inflate groups from OIDC responses
- [x] UI for group discovery and quota control
- [x] Admin-level API for quota usage export


dkfellows:
The problems remaining are now that there's a lot of missing code; there's nothing to set up groups (particularly the part to populate them from the OIDC responses) and there's nothing in the UI to allow any sort of control over groups or really quotas either. But jobs are at least accounted for against their relevant quota.

rowleya:
Note that there probably doesn't need to be a UI for this if there is at least some way to e.g. run a script which might read some group quotas from HBP somewhere and then populate the database with this information.  I suspect this would be better as an API of the service rather than trying to directly interact with the database, just to avoid issues of simultaneous access.

dkfellows:
I can get the organisations that a user is a member of, but not the collabratories. Or maybe I can get that, but I don't get get info for myself because I'm not signed up to any collabs; it's hard to tell!

(Experimental work for this is on `spalloc-server` branch for deployment reasons; to test on this branch, I'd need to sort out the DB migration plan or at least rebuild the DB with the new schema. Which is fiddly.)

rowleya:
You should be able to see if you are a member of any collabs from:
https://wiki.ebrains.eu/bin/view/Collabs/#viewer=true&editor=true&admin=true

You could always create a new one of your own here:
https://wiki.ebrains.eu/bin/view/Collabs?clbaction=create

dkfellows:
I was a member of a lot of collabs. I just needed to add the `roles` scope to my app registration. _This is not documented_ except sort of in passing as a note of what the default roles are in a response on [this page](https://wiki.ebrains.eu/bin/view/Collabs/the-collaboratory/Documentation%20IAM/FAQ/OIDC%20Clients%20explained/1.%20Registering%20an%20OIDC%20client/#HCreatingtheclient). That's ***abysmal.*** 😠 

Of course, adding that scope could only be done by registering a new ID because the registration update code for Keycloak fails and fails silently. This is not documented, but I have determined that it is definitely so. That's _also_ ***abysmal.*** 😠 

Now that I've got that fixed, I am now getting both the organizational and collabratory claims, and will be able to do something sensible with them.

dkfellows:
Deployed, so all bugs to be against #325

PR #440, opened 2022-03-28
Reload FPGA definitions

### LINKED ISSUE #429: Reload FPGA Bit files on error
A non-critical future enhancement request.  Every now and again, the flash memory loses the FPGA configuration data and it has to be reflashed.  When this happens, the FPGA ID flag comes back as 3 (which is why spalloc checks these flags when the machines are being allocated).  Currently if the read fails, the allocation is given up, but an alternative would be to automatically perform the steps that are currently done manually to reload the bit files.

Information:
 - The current bitfiles (*.bit)and bmpc script (bt_load) are here: https://github.com/SpiNNakerManchester/SpiNNaker_hardware_tests/tree/master/firmware
 - The bmpc implementation that loads these is here: https://github.com/SpiNNakerManchester/spinnaker_tools/blob/master/tools/bmpc

dkfellows:
fixes #429 

Note that this is _not_ a replacement for `bmpc`. That has things it does that this code doesn't support.


dkfellows:
Merging this for the improvements to the BMP coordinate system. The reloading code will *not* be enabled, at least for now. (Not deploying until I get my laptop fixed.)

PR #444, opened 2022-04-01
Proxy

### LINKED ISSUE #441: Allow proxying of boards
We should be able to ask spalloc server to put up a proxy (HTTP-based for ease of getting through firewalls?) to our job so that we can interact with boards on the big machine from outside the university despite the boards being on internal IP addresses. This would allow people to, say, use robots with the big machine, provided they can handle the latency.

dkfellows:
An integrated system for setting up proxies to SpiNNaker boards. Uses web sockets (will be the `wss:` basic protocol because we're using HTTPS to establish) to provide connection establishment and basic message framing.

fixes #441

Supported _binary_ protocol (within overall WebSocket framing)

<table border>
<tr>
<th>Name</th>
<th>Request Layout (words)</th>
<th>Response Layout (words)</th>
<th>Meaning</th>
</tr>
<tr>
<td>Open Conn.</td>
<td><table border>
<tr>
<td><code>0</code>
<td>CorID</td>
<td>X
<td>Y
<td>UDP<br>Port
</tr>
</table>
(5 words)
</td>
<td><table border>
<tr>
<td><code>0</code>
<td>CorID</td>
<td>ID
</tr>
</table>
(3 words)
</td>
<td>Establish a UDP socket that will talk to the given Ethernet chip within the allocation (i.e., these are job-local coordinates!) on the given port. The CorId is a correlation ID; it's an arbitrary value from the request that the server will pass back unchanged in the response (that enables matching up requests and responses by the client). Returns an ID (non-zero) that can be used to refer to that connection (only guaranteed unique within overall websocket). <p><p>Note that opening a socket declares that you are prepared to receive messages from SpiNNaker on it, but does not mean that SpiNNaker will send any messages that way. All sockets are bidirectional.</td>
</tr>
<tr>
<td>Close Conn.</td>
<td><table border>
<tr>
<td><code>1</code>
<td>CorID</td>
<td>ID
</tr>
</table>
(3 words)
</td>
<td><table border>
<tr>
<td><code>1</code>
<td>CorID</td>
<td>ID/<code>0</code>
</tr>
</table>
(3 words)
</td>
<td>Close an established UDP socket, given its ID. The CorId is a correlation ID; it's an arbitrary value from the request that the server will pass back unchanged in the response (that enables matching up requests and responses by the client). Returns the ID on success, and zero on failure (e.g., because the socket is already closed).</td>
</tr>
<tr>
<td>Send Msg.</td>
<td><table border>
<tr>
<td><code>2</code>
<td>ID
<td>Raw message bytes...
</tr>
</table>
(2 words plus payload bytes)
</td>
<td>N/A</td>
<td>Send a message to SpiNNaker on a particular established UDP configuration. The raw message bytes (<em>including</em> the half-word of ethernet frame padding) follow the header.<p><p>This is technically one-way, but messages come back in the same format (i.e., a 4 byte prefix to say that it is a message, and another 4 bytes to say what socket this is talking about).</td>
</tr>
</table>

Note that closing the websocket closes all UDP connections beneath it, and that they are _not_ intended to be robust across server restarts. Destroying the job _or_ powering down the boards should cause the websockets in that job to be closed (not yet implemented).

### DIFF
--- a/SpiNNaker-allocserv/pom.xml
+++ b/SpiNNaker-allocserv/pom.xml
@@ -352,6 +352,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 			<artifactId>spring-boot-starter-web</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-websocket</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.security</groupId>
 			<artifactId>spring-security-core</artifactId>
 		</dependency>
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/SpallocProperties.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/SpallocProperties.java
@@ -82,6 +82,8 @@ public class SpallocProperties {
 
 	private CompatibilityProperties compat;
 
+	private ProxyProperties proxy;
+
 	public SpallocProperties(//
 			@DefaultValue("spalloc.sqlite3") File databasePath,
 			@DefaultValue("30s") Duration wait,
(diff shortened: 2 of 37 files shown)


rowleya (approved):
The protocol looks good and should meet most if not all needs (only testing will tell)

dkfellows:
Needed to add correlation IDs to the protocol for OPEN and CLOSE so that I could reliably match up the request and response messages on the client side.

dkfellows:
This is now working. When proxying for a transceiver, remember to do one port for SDP to each board, and another to port 54321 on the root board for booting.

dkfellows:
The hard part was lining up all the bits and pieces so that the websocket could be openable, published, found, and secured (so the job details can be picked up). Once they all pointed in the same right direction at once, it worked; the Spring Boot support for binary websockets is pretty good. (So too is the Python side, once you find a working package and discover what its API really is.)

PR #473, opened 2022-04-19
Timeouts should never get multiplied by 1000

dkfellows:
This was a hold-over from when the code was converted from Python, where we were trying to use the same type of timeout type as there. It didn't work out and wasn't necessary, but this held on and was a bit more reachable than expected.

Since we never _deliberately_ passed a `double` to that call as a timeout, just deleting the method is best.

A few other type and doc corrections/improvements too. And a little refactoring.


dkfellows:
I've cleaned up the mess a bit more. Timeouts are now `int` (with `Integer` supported as delegating overrides only). This does mean that “forever” is now only ever about 25 days (2<sup>31</sup> milliseconds) but that'll do; _typical_ timeouts are a few seconds only.

rowleya (approved):
This could easily explain issues I have seen related to timeouts, so definitely happy to merge!

PR #474, opened 2022-04-21
Handle unreachable BMPs

dkfellows:
One of the things I've seen in testing is that sometimes a BMP is unreachable. I'm not sure why this is the case, but it definitely does happen. When that happens, the board *cannot* be switched on (or off, though that's less critical). A failure to switch the board on due to unreachability needs to trigger a call back into the guts of the allocator to allocate a different set of boards, and to mark the board as out of service with a report that an admin can see. (Without that, the job is in a nasty state and the BMP controller doesn't recover well either as it has a task that it can't succeed at.)

We can't go to production until this is fixed.


rowleya:
There has always been an issue with BMPs going non-responsive.  Note that all the BMPs in a rack are connected over the backplane from the first BMP in the rack.  That first BMP should never be asked to do more than one thing at-a-time, as we have observed previously that this most often results in a crash of that BMP.  I don't think that the BMPs crash much when using the current spalloc once that was resolved, but I could be wrong.  I believe they are also supposed to watchdog themselves back into operation should they ever crash, so a possible solution is to wait for a short while for it to come back.

There is the question as to whether the BMPs on the 12-node board you are currently using for testing have been updated to the latest BMP firmware.  If not, this should probably be done (I would guess a BMP version check on those boards vs. the 1 Million Core machine boards might reveal that).

dkfellows:
This is a BMP going unreachable from the single thread that accesses that particular cabinet & frame.

dkfellows:
Will need to deploy this against the broken configuration to test it.

dkfellows:
This does everything I want now, except emailing notifications when something is taken out of service, and that's because the emailing code is currently very entangled in a different part of the code. Merging it because I'm needing some of the support code in #502

PR #482, opened 2022-04-27
Handle EIEIO proxying

dkfellows:
The live packet gatherer needs a totally different approach. Of course. It wants to use a single EIEIO connection and reprogram the tags on multiple ethernet chips to point to it. That in turn works in a totally different mode of operation of sockets to the way that the proxies are working. Ugh.

Needs two extra operations:
1. `OPEN_UNBOUND` &mdash; Open a socket without binding it. (Requires that we validate the addresses of messages passing by it, but allow messages from any board.)
2. `MESSAGE_TO` &mdash; Send a message on an unbound socket to a given board. We specify the board by X,Y; we could do it by IP address, but that seems wonky given that we don't really expose that elsewhere.

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/SpallocProperties.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/SpallocProperties.java
@@ -40,6 +40,8 @@ import org.springframework.boot.context.properties.bind.DefaultValue;
 import org.springframework.core.io.Resource;
 import org.springframework.validation.annotation.Validated;
 
+import uk.ac.manchester.spinnaker.alloc.model.IPAddress;
+
 /**
  * Spalloc service management properties. These are all intended to be set via
  * the configuration file.
@@ -236,6 +238,8 @@ public class SpallocProperties {
 	/**
 	 * @return Properties relating to the SDP proxying.
 	 */
+	@NotNull
+	@Valid
 	public ProxyProperties getProxy() {
 		return proxy;
 	}
@@ -719,7 +723,6 @@ public class SpallocProperties {
 		 */
 		private OpenIDProperties openid;
 
-		@SuppressWarnings("checkstyle:ParameterNumber")
 		public AuthProperties(//
 				@DefaultValue("true") boolean basic,
 				@DefaultValue("SpallocService") String realm,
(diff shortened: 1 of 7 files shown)


dkfellows:
Err, yes. There's actually a local address bound (set by config).

rowleya:
I can see the issue; it confused me a bit but then I know quite a bit about sockets, so maybe this isn't much of a problem for the average user (and given that our own code will do most of the work anyway it is less critical).  If you did want another name, perhaps ```OPEN_TARGETLESS``` would be an option, or something like that.

rowleya:
One thing in here I am struggling to understand is why the EIEIO connection requires a local host.  Normally if you don't provide one, you get a default listening socket I think that listens on all IP addresses.  I think that on the proxy this is the socket that listens to packets from SpiNNaker.  Is there a reason this needs a definite local host IP to be specified?

dkfellows:
I'm still trying to comprehend the twisted guts of things. 😄 

There's a bunch of places in sPyNNaker that seem to care about the local address/port of the socket (specifically in the `EthernetControlConnection` and the `EthernetCommandConnection`) which are participating in the startup notification process. But maybe what's necessary there is to stand up some sort of local end to the proxied socket? Or maybe not.

I've not yet finished figuring out how to handle the buffer manager. (That might at least be a conventional connection, even if dressed up as EIEIO.)

rowleya:
> There's a bunch of places in sPyNNaker that seem to care about the local address/port of the socket (specifically in the EthernetControlConnection and the EthernetCommandConnection) which are participating in the startup notification process. But maybe what's necessary there is to stand up some sort of local end to the proxied socket? Or maybe not.

Although those have local_host as an argument, I don't think they actually care what it is.  I can see that we pass through the values from the call, but those are almost certainly never actually used, and so can probably be ignored.

>I've not yet finished figuring out how to handle the buffer manager. (That might at least be a conventional connection, even if dressed up as EIEIO.)

This may also be "hangover" code left from the times when we would do live input/output of buffers.  I am going to guess that this can be removed also.

dkfellows:
It's going to take a while to sort this all out. I'm unable to deploy anything right now (because I've not got SSH set up on this system at all). I do know that there's going to be some "interesting" complexity later on, when we want to get out-of-process access for visualisers.

dkfellows:
OK, I'm now pretty sure that things are working, at least enough for this to make it to the main branch. It'll still need deploying and more testing, but I believe the API exposed here is workable.

PR #488, opened 2022-05-12
Low level networking rework

dkfellows:
This gets rid of the explicit use of select and just uses timeouts. Those will be either done with select under the covers, or they'll be passed down to the OS (perhaps with `SO_RCVTIMEO`) but that's the Java implementation's problem. The good thing about this is that it simplifies the low level code underneath connections quite a bit. (The bad thing is that manipulating buffers is now messier.)

We also specify the traffic class; it's tuned for throughput on the types of connections where we're most likely to need to move a lot of data.


dkfellows:
I believe this is incompatible with #482 but I'll just fix that PR when needed.

rowleya (approved):
This looks fine.  I ran some integration tests to be confident and nothing of concern seems to be raised.

PR #489, opened 2022-05-13
Rework the timeout logic for fast connections

### LINKED ISSUE #453: Timeout doesn't get triggered on download
This was run with the reduce_overhead branch but I don't think the code differs in any important way here from master.  The code freezes on download.  The jspin log is apparently too big to attach, but a jstack has been output and attached.

[jstack.txt](https://github.com/SpiNNakerManchester/JavaSpiNNaker/files/8462808/jstack.txt)

dkfellows:
The aim is that there's about 30 seconds (a bit over) after the last successful receive from a core before we time out the connection; any successful receive resets the counter. This lets use a shorter count of retries. (Previously, the timeout was effectively in the region of 200 seconds... if no messages were ever received, and infinite if one was received; a crashing core would make things go very badly wrong.)

I hope this fixes #453


rowleya (approved):
This looks fine also.  I was a bit worried about getting more timeouts in general, but I can see now that this shouldn't happen.

PR #495, opened 2022-05-19
Java 11

dkfellows:
We're starting to see the removal of support for Java 8, mostly in dependencies (manifesting as incompatible bytecode versions). We're going to have to move...

- [x] get changes right.
- [ ] update expected builds to require the correct set of builds.
- [ ] ensure that we have Java 11 on the target machines.


dkfellows:
Obsoleted by #509

PR #501, opened 2022-05-24
Allow virtual chips outside of the size of a board

rowleya:
Avoids an error with virtual chips on 4-node boards


Christian-B:
Ok to merge with my fiddle

PR #502, opened 2022-05-25
Manage hardware blacklists

### LINKED ISSUE #500: Support updating of blacklist files?
A bit related to #429, I was wondering if it is also add an interface that would let admins blacklist cores and/or chips on the machine.  This should just require a BMP write of some sort, though some details would likely be required.

Some details how how this is currently done are here:
https://github.com/SpiNNakerManchester/SpiNNaker_hardware_tests/tree/master/blacklist_files

dkfellows:
fixes #500


dkfellows:
Well, we've now got the blacklist reading code, though the blacklist data might not be rendered in the most elegant way yet. Assuming that the code is working.

The blacklist writing is not hooked up. The machinery is there to push the blacklist from the DB to the right BMP, but I'm not sure how we want to go about editing blacklists. Simple text is much easier than a fancy GUI.

rowleya:
Email notifications for automatic removal would definitely be nice to have; for manual removal this is much less critical also.

dkfellows:
Disentangling the email sender is... necessary but awkward.

dkfellows:
Main thing missing now is a mechanism for editing a blacklist. Doing it as text would be easy (I already have parsers and text renderers), but a GUI would allow the validation that was being talked about in #500. The problem with a full GUI is that it is quite a lot more work!

dkfellows:
> admins should know what they are doing

I'll quote you on that. 😏

dkfellows:
Main thing left to do is to write a test that proves we can route blacklist read and write messages from the `MachineStateControl` to the `SpiNNaker1` and back, which is a complex test (really an integration test, but one that can run during the unit test phase as it can be isolated from real hardware).

dkfellows:
I believe this will now work.

Note that it will be a pain to deploy the first time after this; the database will need to be *recreated* because of the way that the schema has changed (mostly to do with the changes to how intra-board descriptions work, which was poorly modelled before).

PR #509, opened 2022-06-15
Update to use Java 11 

dkfellows:
This is #495 for post-#325

- [x] get changes right.
- [x] update expected builds to require the correct set of builds.
- [x] ensure that we have Java 11 on the target machines.


dkfellows:
Discussion in the team leads me to conclude we'll be going with this fairly soon.

PR #515, opened 2022-06-27
Reduce overhead

rowleya:
Some things from getting the multiarea model working including:
1) Timeouts don't need to be Integers
2) Better handling of receive with IP address (appears to work where old method didn't)
3) Calculate checksums (Java equivalent of SpiNNakerManchester/DataSpecification#121)
4) Avoid missing region reference errors
5) Log unexpected command code, but don't break because of it (it appears to be recoverable)


rowleya:
> What is the benefit of your `UDPPacket` over the standard `DatagramPacket`?

The only advantage is that it can contain the same "sort of stuff" that we already handle elsewhere, so we don't need to do different things depending on whether we include the IP address or not.

SpiNNaker-data-specification/src/main/java/uk/ac/manchester/spinnaker/data_spec/Executor.java
```diff
@@ -208,7 +211,30 @@ public void addPointerTable(ByteBuffer buffer) {
...
+							/ INT_SIZE);
+					ByteBuffer bytebuf = (ByteBuffer) regReal.getRegionData()
+							.duplicate().order(LITTLE_ENDIAN).rewind();
+					IntBuffer buf = bytebuf.asIntBuffer();
+					long sum = 0;
+					for (int i = 0; i < nWords; i++) {
+						sum = (sum + (buf.get() & UNSIGNED_INT)) & UNSIGNED_INT;
+					}
+					// Write the checksum and number of words
+					buffer.putInt((int) (sum & UNSIGNED_INT));
```
dkfellows, line 227: Is this complexity necessary? Java, unlike C, has defined `int` overflow rules.

dkfellows:
> The only advantage is that it can contain the same "sort of stuff" that we already handle elsewhere

Yes... but the stuff it is holding is pretty much the same as is in the standard one. (I'm reluctant to set the endianness at the base network level because that's very strongly protocol dependent, and we use a few different protocols...)

SpiNNaker-data-specification/src/main/java/uk/ac/manchester/spinnaker/data_spec/Executor.java
rowleya: From memory, doing it with a simple sum fails.  My assumption is that this is due to the unsigned-ness.  The C code does unsigned addition to find the same sum, so the worry would be that making this a signed sum might make that fail (bearing in mind that each item in the buffer might be "positive" or "negative" since it is just data).

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/BufferManagerStorage.java
```diff
@@ -192,6 +192,11 @@ public final boolean isAligned() {
 		public final boolean isNonEmpty() {
 			return size > 0;
 		}
+
+		@Override
+		public String toString() {
+			return "Region " + regionIndex + " on " + core.toString();
```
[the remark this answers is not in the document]
rowleya: True; I have tended to use this just to be clear in a toString.  As you say, it doesn't technically matter since that is what will happen anyway!

dkfellows:
The differences between a standard packet and the one you've specified are probably either in the two prefix bytes or the endianness.

It isn't so important now that we are using the old `DatagramSocket` classes.

dkfellows (approved):
If this works, it can go in when the rest of the changeset is ready.

PR #526, opened 2022-07-12
Make memory locations into their own type

dkfellows:
There were too many methods in the transceiver still taking multiple `int`s as arguments despite there being many different meanings to them. Using a custom type for the memory location resolves that. It also gets rid of the signed/unsigned problem; it's a memory location! (Internally it uses standard unsigned handling methods when appropriate.)


dkfellows:
A side benefit of this is that we get rid of all those duplicated methods that took a `long` instead of an `int`. `MemoryLocation` is effectively a *tagged, unsigned* `int` that renders to hexadecimal always, and handles everything relating to that correctly. The only places where we look at its `address` field from the outside of the class should be where we're feeding it into some API that can't really accept a `MemoryLocation`, typically because it is being fed into a `ByteBuffer` to be sent to SpiNNaker...

PR #531, opened 2022-07-22
A different way to do short waits

dkfellows:
If this works, it will be better than the busy loop it replaces. It's *supposed* to sleep using the high-performance system calls (high-resolution timers, etc.) for doing so.

The referenced question is [this one](https://stackoverflow.com/questions/35875117/can-a-thread-sleep-for-less-than-half-milli-seconds-in-java-other-language), but the `LockSupport` class mentioned in the accepted answer is pretty obscure (despite being available since Java 5).


dkfellows:
Did a force-update. Neater than a merge of `master` in this case.

dkfellows:
This particular PR _must_ be tested before it goes in. Luckily, it is nicely isolated from everything else and so isn't onerous to keep up to date.

PR #535, opened 2022-07-27
Modularise the code

dkfellows:
Adds module-info.java files that define what's the public API and what's just the implementation. JPMS support is non-trivial, and may require further work.

For more info, see [this tutorial](https://www.baeldung.com/java-9-modularity) which cuts out a lot of the confusing parts.


dkfellows:
This appears to work. I don't want to merge it yet though; want to reconsider what classes are in what packages (especially for internal classes that shouldn't be used by other code but which need to be `public` anyway).

dkfellows:
In theory, `SpiNNaker-comms` should be reorganised so that it doesn't export so much.

In practice, it's got far too many subtle inter-package dependencies.

dkfellows:
Parking this for now. Until Java, Maven, JUnit, and Eclipse can conspire to let me test my code (without using workarounds that one or another part can't handle) then I'm not going to make any progress.

dkfellows:
This is a mess and going to remain so. Closing. *After the update to Spring 6* (#651) I might think about redoing this. If Spring is modularised; there's no point in trying until that's sorted out.

PR #559, opened 2022-08-23
More javadoc finickiness

dkfellows:
This PR is a consequence of me using `mvn javadoc:aggregate` to find publishing issues and other annoying things like that.

We stop documenting `keycloak-management-client` because it was causing the oddest failures in doc builds and is really not something that users can use. (You have to edit the code to make changes at all; it's extremely non-general!)

### DIFF
--- a/SpiNNaker-allocator/src/main/java/uk/ac/manchester/spinnaker/allocator/AllocatedMachine.java
+++ b/SpiNNaker-allocator/src/main/java/uk/ac/manchester/spinnaker/allocator/AllocatedMachine.java
@@ -52,6 +52,7 @@ public class AllocatedMachine {
 		return width;
 	}
 
+	/** @param width Rectangle width. */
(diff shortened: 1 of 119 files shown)


dkfellows:
In some cases, we "document" things by changing the visibility of some setters to package-private. If the class is definitely never meant to have those called except via Jackson (which can see behind the curtain) then it doesn't need to wave them around.

The change is large simply because it documents *all* those little getters. And I need to double check that everything used in the UI still has enough visibility because those are pseudo-scripted.

rowleya:
I believe that Jackson can set private properties directly if you so wish i.e. no need for a setter...

dkfellows:
Checks:

- [x] `SystemControllerImpl`
- [ ] `AdminControllerImpl`
- [ ] `RootController`

dkfellows:
Looks OK. Or rather #560 isn't a bug with this branch.

PR #569, opened 2022-08-31
Proxied transceiver for Java spalloc client

### LINKED ISSUE #551: Augment SpallocClient.Job with getTransceiver()
The Java client for new Spalloc should support getting a (proxied) transceiver, just as the equivalent Python code does.

dkfellows:
This is about enhancing the Java spalloc client so that it can create proxied transceivers just as the Python one can do.

This fixes #551 eventually.


dkfellows:
Just needs documenting now. I'll postpone implementing proxied sockets that need sendto/recvfrom until they're actually needed.

dkfellows:
I implemented the unconnected sockets. It turned out they were easy because we were already setting up EIEIO sockets to work in a similar way.

PR #575, opened 2022-09-06
Login with bearer token

dkfellows:
I don't know if this works. I know we want to establish an authenticated session, and I know that I'm performing an action with a session where I provide the bearer token at the point when I would otherwise supply the username and password, but I really don't know if the other side will like this at all. It might need a different URL to do the proper exchange of long-term credentials for a session token...


rowleya (approved):
This looks correct enough.  Note that a bearer token is a temporary thing that the user must obtain by logging in on the web normally, so getting one might not be easy...  A possible other option would be to support a local token authentication scheme where a user logged in to the web interface in any existing way (including via EBRAIN OIDC) can create a local token, which can then be used to authenticate that user until revoked.

PR #576, opened 2022-09-07
Fix the FIXME

dkfellows:
The method was never called, and only had a dummy implementation and the last remaining `FIXME` comment. It can simply go.


dkfellows:
Well that was messy! Should be good now.

PR #579, opened 2022-09-14
Reduce duplicates

dkfellows:
Let's try to have fewer places where we duplicate code. That'd be nice.


dkfellows:
This branch is a disaster now; too difficult to merge. I'll recover the interesting bits in new PRs.

PR #581, opened 2022-09-16
Small changes

dkfellows:
* Avoid pointless null checks (instanceof does that too)
* Add final to classes intended to be used as hash table keys
* Add javadoc


dkfellows:
The `Objects` class has so many *useful* little utilities.

PR #602, opened 2022-09-30
Reduce duplicates: validation

dkfellows:
Move validator to machine and add one for UDP port numbers. Still need to apply these where appropriate.

Recovered from #579


dkfellows:
I *hope* that I've got the range of valid values for triad coordinates correct; it's not very well documented! (The valid range of chip coordinates is simpler; X and Y have to each fit into a byte.)

dkfellows:
That's got validation annotations done on lots of things. Whether things take advantage of them or not, they're there for when they decide to use them.

PR #609, opened 2022-10-06
Reduce duplicates

dkfellows:
Recovery of main idea from #579.


dkfellows:
I've decided to skip the recovery of the bits where I was trying to unify the notions of directions. That's just too ghastly. I'll leave it the way it was; that at least works even if it is sometimes confusing.

PR #614, opened 2022-10-11
Add in explicit dependency resolution step

dkfellows:
The output in build logs when downloading dependencies is awful. This PR tries to put that all in its own build stage (in batch mode) so that that noise isn't spread everywhere. 

**TRICKY POINT:** extra maven profiles have to be activated when doing this so dependencies in those steps are also found.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,10 +41,10 @@ jobs:
         java-version: ${{ matrix.java }}
         cache: 'maven'
 
-    - name: Resolve Dependencies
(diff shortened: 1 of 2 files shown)


dkfellows:
Ugh. It doesn't exactly work right because [MDEP-830](https://issues.apache.org/jira/browse/MDEP-830) is in the way. This is pitiful, but not easily ~resolved~ sorted out.

dkfellows:
That's better. The concept is on the build workflow, which I don't like, but the details are just in the POM and that's much better.

PR #620, opened 2022-10-17
Add InterruptedException to general transceiver operation profile

dkfellows:
While tidying up, I noticed that there was interrupts of waiting in the guts of the transceiver that was getting swallowed. Oh dear! Stopping that from vanishing nastily (i.e., letting the `InterruptedException` do what it should) caused the transceiver's public exception profile to change, and that in turn cascades into many places.


dkfellows:
This was not originally in #579, but fits with that PR's philosophy.

PR #634, opened 2022-11-03
Move new spalloc client to SpiNNaker-comms

dkfellows:
The new client's best in with the old one as they're both pretty close to the comms layer now (especially the new one, with its proxying support). This will enable making things better integrated.

*This* PR does *not* clean up the internals of the new client or alter any Java package names. It just rearranges the code for the purposes of building.


dkfellows:
Known to interfere with #575

PR #636, opened 2022-11-03
Rearrange exceptions

dkfellows:
Move many exceptions (which are generally pretty boring in themselves, even if what they describe isn't) into their own Java packages. Try to make some of the `ProcessException` throwing places give more information out so that we can better figure out what's failing when things go wrong.


dkfellows:
Noisy, but not really valuable.

PR #641, opened 2022-11-07
Fix for failing tests

dkfellows:
I'd got some logic wrong when factoring out slicing. The slicing itself was fine, but the call was wrong. Add a test of slicing


Christian-B (approved):
As master is more broken that this I suggest merging and then reruning Jenkins on master

PR #643, opened 2022-11-15
More javadoc links

dkfellows:
This is all about getting better cross-references to classes we use (especially annotation types and JAX-RS bits).

It also turns a few things in builders back into setters; the `withFoo()` fluent style encouraged by Jackson doesn't really do it for me.


dkfellows:
I'm still seeing some bugs in generated docs when I run locally, but they're faults in Javadoc (especially related to annotated parameters with primitive types) and I'm not about about to fix those!

dkfellows:
The doc build now produces warnings about redirects. I don't care; I want the URLs I used to end up in the docs because that's better in other ways. Stupid tools...

PR #655, opened 2022-11-23
Return more details of the error to the user

rowleya:
Extends the response to the user in the event of an error


dkfellows:
The tricky bit is working out whether the information that is sent back to the user is information that they should be told. Which is obvious except, haha, it isn't. Review will take a bit of time for this because I need to trace error paths.

rowleya:
> The tricky bit is working out whether the information that is sent back to the user is information that they should be told. Which is obvious except, haha, it isn't. Review will take a bit of time for this because I need to trace error paths.

All I know is that this told me that the user didn't have any quota, which was much better than just a generic error message.  Happy for more tracing to be done though...

dkfellows:
Lack of quota ought to be handled one way... but a failure to map the group would have got lost/generic-ed. Oops!

dkfellows (approved):
I still worry about what random thing might show up, but at least I'm sure this won't leak stack traces or anything else like that that really ought to never make it to users.

PR #656, opened 2022-11-24
Enhance the OAuth2 bearer token we get 

dkfellows:
The JWT doesn't always include all the claims we need. Specifically, it sometimes misses the `roles/team` sub-claim. To fix this, we need to query Keycloak to get the extra claims info we need.

Try to do auth using the (opaque) token introspection scheme, which apparently works just fine with JWT too. If this doesn't work (it depends on *exactly* what the server sends back and it's all extremely soft coded), we'll need to poke the `userinfo` endpoint.


dkfellows:
#657 is probably better.

PR #657, opened 2022-11-24
Fixes to make bearer tokens do the right thing

rowleya:
A way to make it work...

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/security/SecurityConfig.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/security/SecurityConfig.java
@@ -25,33 +25,55 @@ import static uk.ac.manchester.spinnaker.alloc.security.Utils.loadTrustStore;
 import static uk.ac.manchester.spinnaker.alloc.security.Utils.trustManager;
 
 import java.io.IOException;
+import java.net.URI;
 import java.security.GeneralSecurityException;
+import java.time.Instant;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Map;
 
 import javax.net.ssl.X509TrustManager;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


SpiNNaker-allocserv/src/main/resources/application.yml
```diff
@@ -155,6 +155,8 @@ spalloc:
       # Support OIDC
       enable: true
       domain: https://iam.ebrains.eu/auth/realms/hbp
+      introspection: https://iam.ebrains.eu/auth/realms/hbp/protocol/openid-connect/token/introspect
+      userinfo: https://iam.ebrains.eu/auth/realms/hbp/protocol/openid-connect/userinfo
```
dkfellows, line 159: These two entries ought to have corresponding bits in `SpallocProperties`. There's a few reasons for that, but the core one is that it allows us to "typecheck" the properties (in this case making sure that they work as URLs, even if we don't &mdash; and *won't* &mdash; make sure that they exist during startup).

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/security/SecurityConfig.java
```diff
@@ -154,6 +183,18 @@ static X509TrustManager customTrustManager(AuthProperties props)
 	@Autowired
 	private URLPathMaker urlMaker;
 
+	@Value("${spalloc.auth.openid.introspection}")
```
dkfellows, line 186: This should be obtained from the already-autowired `SpallocProperties.AuthProperties` instance.

dkfellows (changes requested):
Needs to plug into the general configuration system (`SpallocProperties`), which is used to provide more discoverability in the UI (it's magic involving a hidden preprocessor) and general typechecking and basic validation of configuration.

I can sort this out

dkfellows (approved):
If these work (and don't disrupt logging in with openid on the web UI, which I don't think they should) then they're fine now.

PR #658, opened 2022-11-28
Update Java version to 17 and JEE to 9

dkfellows:
This is the migration process where we adopt new language and library features, as well as updating the JEE profile. We're not stopping off at 14 on the way because Spring and Jakarta are going to force the larger stride.

The update to JEE 9 includes Spring/Spring Boot 6 and CXF 4. Those _require_ Java 17.


dkfellows:
**Merging instructions:** prior to merging this PR, you need to change what the expected set of passing tests is in the Repository Settings for the `master` branch. This PR _removes_ support for Java 11 and 14, and _adds_ the need for Java 17 to pass for both testing and validation.

rowleya:
Should have spotted this sooner... updates now done elsewhere!

PR #664, opened 2022-12-01
Spalloc server general fixes

rowleya:
General fixes that make the spalloc client end work, as well as actual use of spalloc proxies in the Front End code.

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -353,7 +353,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 	@Override
 	@PostFilter(MAY_SEE_JOB_DETAILS)
 	public Optional<JobDescription> getJobInfo(Permit permit, int id) {
-		return execute(conn -> {
+		return executeRead(conn -> {
 			try (var s = conn.query(GET_JOB);
 					var chipDimensions = conn.query(GET_JOB_CHIP_DIMENSIONS);
 					var countPoweredBoards = conn.query(COUNT_POWERED_BOARDS);
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -1151,6 +1151,24 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 			partial = false;
 		}
 
+		private MachineImpl cachedMachine;
+
+		/**
+		 * Get the machine that this job is running on. May used a cached value.
+		 * A transaction is required, but may be a read-only transaction.
+		 *
+		 * @param conn
+		 *            The connection to the DB
+		 * @return The overall machine handle.
+		 */
+		private synchronized MachineImpl getJobMachine(Connection conn) {
+			if (cachedMachine == null || !cachedMachine.epoch.isValid()) {
+				cachedMachine = Spalloc.this.getMachine(machineId, true, conn)
+						.orElseThrow();
+			}
+			return cachedMachine;
+		}
+
 		@Override
 		public void access(String keepaliveAddress) {
 			if (partial) {
@@ -1247,8 +1265,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 				return conn.transaction(false, () -> findBoard
 						.call1(id, root, x, y)
 						.map(row -> new BoardLocationImpl(row,
-								Spalloc.this.getMachine(machineId, true, conn)
-										.orElseThrow())));
+								getJobMachine(conn))));
 			}
 		}
 
@@ -1452,8 +1469,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 			private List<Integer> boardIds;
 
 			private SubMachineImpl(Connection conn) {
-				machine = Spalloc.this.getMachine(machineId, true, conn)
-						.orElseThrow();
+				machine = getJobMachine(conn);
 				try (var getRootXY = conn.query(GET_ROOT_COORDS);
 						var getBoardInfo = conn.query(GET_BOARD_CONNECT_INFO)) {
 					getRootXY.call1(root).ifPresent(row -> {
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -130,7 +130,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 
 	@Override
 	public Map<String, Machine> getMachines(boolean allowOutOfService) {
-		return execute(c -> getMachines(c, allowOutOfService));
+		return executeRead(c -> getMachines(c, allowOutOfService));
 	}
 
 	private Map<String, Machine> getMachines(Connection conn,
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -176,7 +176,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 	public List<MachineListEntryRecord>
 			listMachines(boolean allowOutOfService) {
 		try (var sql = new ListMachinesSQL()) {
-			return sql.transaction(false,
+			return sql.transactionRead(
 					() -> sql.listMachines.call(allowOutOfService)
 							.map(sql::makeMachineListEntryRecord).toList());
 		}
@@ -242,7 +242,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 	public Optional<MachineDescription> getMachineInfo(String machine,
 			boolean allowOutOfService, Permit permit) {
 		try (var sql = new DescribeMachineSQL()) {
-			return sql.transaction(false, () -> apply(
+			return sql.transactionRead(() -> apply(
 					sql.namedMachine.call1(machine, allowOutOfService)
 							.map(Spalloc::getBasicMachineInfo),
 					md -> sql.countMachineThings.call1(md.getId())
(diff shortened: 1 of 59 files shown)


SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/transceiver/Transceiver.java
```diff
@@ -2177,9 +2177,16 @@ public void setIPTag(IPTag tag, SDPConnection connection)
 					"The given board address is not recognised");
 		}
 
-		simpleProcess(connection)
-				.synchronousCall(new IPTagSet(connection.getChip(), null, 0,
-						tag.getTag(), tag.isStripSDP(), true));
+		// Avoid delegation of the process if not needed
```
dkfellows, line 2180: It seems to me like this ought to be something that gets wrapped into `simpleProcess()` or at least into a helper method.

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/ProxyInformation.java
```diff
@@ -0,0 +1,57 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package uk.ac.manchester.spinnaker.storage;
+
+import javax.validation.Valid;
+
+/**
+ * Information about the proxy to allow connection.
+ */
+public class ProxyInformation {
```
dkfellows, line 24: The validity definitions in this class need to change. `@Valid` says "use the validity rules from the class of thing it is" and `String` doesn't have any as it is a built-in class.

I'd guess that `@NotBlank` would be a good start; tmpty strings are definitely not going to work.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
```diff
@@ -106,10 +108,17 @@ public class SpallocClientFactory {
...
 			.findAndAddModules().disable(WRITE_DATES_AS_TIMESTAMPS)
+			.addModule(new JavaTimeModule())
 			.propertyNamingStrategy(KEBAB_CASE).build();
 
 	private final URI baseUrl;
 
+	/**
+	 * Cache of machines, which don't expire.
+	 */
+	private static final Map<String, Machine> MACHINE_MAP =
```
dkfellows, line 119: Formally, this could get into trouble if you were writing code that talks to multiple spalloc servers at once where those have machines with the same name. It's not a scenario that I think is going to be very worrying for anyone other than ourselves.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/ClientSession.java
```diff
@@ -673,11 +661,14 @@ private void logSessionIn(String tempCsrf) throws IOException {
 	 *             If things go wrong.
 	 */
 	private synchronized void renew(boolean postRenew) throws IOException {
+
 		// Create a temporary session so we can log in
 		var tempCsrf = makeTemporarySession();
 
-		// This makes the real session
-		logSessionIn(tempCsrf);
+		// If we didn't use a bearer token, we need to log in properly
```
dkfellows, line 668: With a bearer token, is the initial session authorized then? I was wondering about that. What we don't want to do is to send the bearer token every time, since that forces a (slow) roundtrip to keycloak on every service call.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/ClientSession.java
```diff
@@ -580,21 +583,13 @@ private synchronized void authorizeConnection(HttpURLConnection c,
...
-				continue;
-			}
-			var setCookie = conn.getHeaderField(i);
-			if (setCookie != null) {
-				var m = SESSION_ID_RE.matcher(setCookie);
-				if (m.find()) {
-					session = m.group(1);
-					found = true;
-				}
+		var setCookie = conn.getHeaderField(SET_COOKIE);
```
dkfellows, line 586: I didn't use this method because there can be *multiple* `Set-Cookie` headers in a response (it's one per cookie being set *or deleted* &mdash; I know that's not formally part of the spec but it damn well exists in practice!) and I can't be certain that there aren't more than one or what order they are.  The javadoc string for the one you're using says:

     * Returns the value of the named header field.
     * <p>
     * If called on a connection that sets the same header multiple times
     * with possibly different values, only the last value is returned.

So yes, it'd make the code neat if we could use it, and conceptually it's what we're doing, but I suspect we can't do so safely.

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/sqlite/SQLiteProxyInformation.java
```diff
@@ -0,0 +1,76 @@
...
+	 * @param conn The connection to read the data from.
+	 * @return The proxy information.
+	 * @throws SQLException If there is an error reading the database.
+	 */
+	public static ProxyInformation getSQLProxyInformation(Connection conn)
+			throws SQLException {
+		String spallocUri = null;
+		String jobUri = null;
+		String bearerToken = null;
+		try (var s = conn.prepareStatement(GET_PROXY_INFORMATION)) {
```
dkfellows, line 49: This line can be combined with the next one, and try-with-resources will do the Right Thing&trade;. It's neat that works.

dkfellows (changes requested):
A few things I'm a touch concerned about.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/transceiver/Transceiver.java
rowleya: Fixed.

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/ProxyInformation.java
rowleya: Sorry didn't look at this in great detail!  Added notempty for now...

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
rowleya: I am not overly concerned with this!

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/ClientSession.java
rowleya: Yes the initial session seems to be authorized once the bearer is used.  It still gets the cookie here so it seems to work...

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/ClientSession.java
rowleya: Oh yes... it wasn't working as it was, but I have updated it now so it should work...

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/sqlite/SQLiteProxyInformation.java
rowleya: Ah yes... I know about these things, just that they don't fit into my current way of thinking!  Fixed now.

dkfellows (approved):

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
```diff
@@ -130,7 +130,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 
 	@Override
 	public Map<String, Machine> getMachines(boolean allowOutOfService) {
-		return execute(c -> getMachines(c, allowOutOfService));
+		return executeRead(c -> getMachines(c, allowOutOfService));
```
dkfellows, line 133: This was another wrong (over-locked) transaction type.

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/bmp/BMPController.java
```diff
@@ -162,6 +164,13 @@ public class BMPController extends DatabaseAwareBean {
 	@GuardedBy("this")
 	private Throwable bmpProcessingException;
 
+	/**
+	 * What jobs are currently being processed? Matters because they might be
+	 * busy because of BMPs that are busy with reloading FPGAs, which is a very
+	 * slow business because bandwidth isn't great on the management channel.
+	 */
+	private final Set<Integer> busyJobs = new ConcurrentSkipListSet<>();
```
dkfellows, line 172: This is kept only in memory because restarting the server invalidates any information that it might have.

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/db/DatabaseAwareBean.java
```diff
@@ -149,22 +149,18 @@ public final <T> T transaction(TransactedWithResult<T> action) {
...
-		 *            <em>cannot</em> be upgraded (because that causes
-		 *            deadlocks).
 		 * @param action
 		 *            The code to run inside the transaction.
 		 * @return Whatever the {@code action} returns.
 		 */
-		public final <T> T transaction(boolean lockForWriting,
-				TransactedWithResult<T> action) {
-			return conn.transaction(lockForWriting, action);
+		public final <T> T transactionRead(TransactedWithResult<T> action) {
```
dkfellows, line 162: Less error prone to do it like this.

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/db/Utils.java
```diff
@@ -99,8 +98,33 @@ private static String trimSQLComments(String sql) {
...
+
+	/**
+	 * Utility for testing whether an exception was thrown because the database
+	 * was busy.
+	 *
+	 * @param exception
+	 *            The exception to test.
+	 * @return Whether it was caused by the database being busy.
+	 */
+	public static boolean isBusy(SQLException exception) {
```
dkfellows, line 115: Because writing this `switch` several times gets old...

SpiNNaker-front-end/src/main/java/uk/ac/manchester/spinnaker/front_end/dse/ThrottledConnection.java
```diff
@@ -101,16 +103,21 @@ class ThrottledConnection implements Closeable {
...
+	@SuppressWarnings("MustBeClosed")
 	ThrottledConnection(TransceiverInterface transceiver, Ethernet board,
-			IPTag iptag)
+			IPTag iptag, SpallocClient.Job job)
 			throws IOException, ProcessException, InterruptedException {
 		location = board.location;
-		addr = getByName(board.ethernetAddress);
-		connection = new SCPConnection(location, addr, SCP_SCAMP_PORT);
+		if (job == null) {
+			connection = new SCPConnection(location,
```
dkfellows, line 112: Future TODO: get the transceiver to manufacture the connection? That would let it hide whether it is using the proxy, and also the transceiver already knows the machine layout. Hmm.

Not worth holding back this PR.

PR #684, opened 2023-01-11
Workaround for weird reordering with @Mixin

dkfellows:
The picocli `@Mixin` annotation was making things be reordered in a way that was surprising to me! Switching to `@ArgGroup(multiplicity = "1")` stops that, but is a less-than-obvious alternative given how things are documented.

Reported upstream: https://github.com/remkop/picocli/issues/1919


rowleya (approved):
This appears to pass testing now

remkop:
> I _think_ I worked out what the problem was. For reasons best known to itself, picocli was sorting the positional parameters by their arity.
> 
> I can't even...

Oh I see now! Sorting by arity for positional parameters does not sound great indeed. That would have been a useful piece of information to report in https://github.com/remkop/picocli/issues/1919 ...
I will take a look.

PR #685, opened 2023-01-12
Reliable test job

dkfellows:
One of our tests was disabled because it was not generating reliable results (and had become steadily more broken over time; code that isn't tested at all is code that can be going horribly wrong). This is about trying to make it run and be more reliable.


dkfellows:
Reliability increased by:

1. ensuring that worker threads in the mock server have actually started
2. defining and using a connection object (wrapping the connected socket) so that a server can have multiple connections open at once
3. making the match for the number of keepalives less precise; the exact number is *not* important, so long as they're sent.

dkfellows:
I've not yet removed this comment:
```java
	/*
	 * TODO test need fixing as calls to client are time dependent so this test
	 * sometimes fails. For example j.getState() may or may not call the client
	 * depending on when the previous state was requested.
	 */
```
I'm not sure if it applies or not.

dkfellows:
That failure up above was a transient in networking unrelated to our tests.

dkfellows:
If this causes failures, I'll be very sorry and fix it then.

PR #686, opened 2023-01-13
Remove the unused io package

dkfellows:
The `uk.ac.manchester.spinnaker.io` package was unused. It was modelling an approach in SpiNNMan that wasn't being used, and was even less likely to be used from Java as it wasn't emulating Java I/O streams properly. Removing it also means that we get rid of slices; they're a concept not used otherwise.


rowleya (approved):
Yes, if this is not used now I can't see it being used in the future

PR #692, opened 2023-01-17
Add extension point for SCPConnection making to TransceiverInterface

### LINKED ISSUE #671: Transceiver should manufacture SCPConnections
Should we get the transceiver to manufacture the `SCPConnection` (or its proxy)? That would let it hide whether it is using the proxy, and also the transceiver already knows the machine layout and so wouldn't need quite so much information passing in at this point. Hmm.

_Originally posted by @dkfellows in https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/664#discussion_r1049880769_

dkfellows:
This lets us reduce the amount of need for keeping knowledge of jobs within the front end; the transceiver has that info and makes proxies if it makes sense to do so. Getting that right involves pushing some more smarts into `SCPConnection` so that we don't need to have that level of smartness in the users of it (it has subclasses that override the smartness).

This fixes #671


dkfellows:
I'm also re-synchronizing a file that was not synchronized, removing duplicated fields (where both the superclass and the subclasses had the same field with the same name!) and removing a bunch of constructors that simply weren't used (and which delegated to other constructors, as a way to "add default arguments" which we don't need except in the most common cases).

PR #694, opened 2023-01-19
Correct use of GROUP BY

### LINKED ISSUE #693: Undefined value for bare column sql group by
In some of the queries of the allocator a "group by" clause is used with "bare columns" (see https://www.sqlite.org/lang_select.html section 2.5).  This will lead to potentially incorrect results when e.g. selecting a rectangle starting at a specified coordinate or finding all possible rectangles including those with wrap arounds.

dkfellows:
This makes the code use `GROUP BY` and aggregates correctly, not relying on the DB to select a special row from the group for us. We simply can't be sure that the DB will choose correctly. Fortunately, we have indices already set up that let us efficiently handle this by using the coordinates of the root of the group (which we were grouping by) and inflating that back to the ID of the board at that point.

The report handling just looked wrong.

This fixes #693

### DIFF
--- a/SpiNNaker-allocserv/src/main/resources/queries/find_rectangle_at.sql
+++ b/SpiNNaker-allocserv/src/main/resources/queries/find_rectangle_at.sql
@@ -18,43 +18,36 @@ WITH RECURSIVE
 	args(root_id, width, height, machine_id, max_dead_boards) AS (
 		VALUES (:board_id, :width, :height, :machine_id, :max_dead_boards)),
 	-- Profile the machines and boards to the one we care about
-	m AS (SELECT machines.* FROM machines JOIN args USING (machine_id) LIMIT 1),
-	bs AS (SELECT boards.* FROM boards JOIN args USING (machine_id)),
+	m AS (
+		SELECT machines.* FROM machines JOIN args USING (machine_id)
+		LIMIT 1),
+	bs AS (
+		SELECT boards.* FROM boards JOIN args USING (machine_id)),
+	selected_root AS (
+		SELECT bs.* FROM bs JOIN args ON bs.board_id = args.root_id
+		LIMIT 1),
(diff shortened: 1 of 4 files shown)


SpiNNaker-allocserv/src/main/resources/queries/find_rectangle_at.sql
```diff
@@ -32,19 +32,29 @@ WITH RECURSIVE
 		LIMIT (SELECT height FROM m)),
 	-- Form the sequences into grids of points
 	c(x,y,z) AS (SELECT x, y, z FROM cx, cy, triad),
-	g(x,y) AS (SELECT x, y FROM gx, gy)
+	g(x,y) AS (SELECT x, y FROM gx, gy),
+	-- Root coords and number of boards available from that point
+	-- NB: Can't use board ID safely as we are using a GROUP BY
+	root(x,y,z,available) AS (
+		SELECT g.x AS x, g.y AS y, 0 AS z,
+			SUM(boards.may_be_allocated) AS available
```
rowleya, line 40: I think this needs to be bs.may_be_allocated

SpiNNaker-allocserv/src/main/resources/queries/get_reported_boards.sql
```diff
@@ -14,17 +14,22 @@
 -- along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 -- Get boards with more problem reports than a critical threshold
+WITH report_counts AS (
+	SELECT
+		board_reports.board_id,
+		COUNT(board_report.report_id) AS num_reports
```
rowleya, line 20: This should be board_reports.report_id

rowleya (changes requested):
Couple of fixes to do

rowleya (approved):

PR #695, opened 2023-01-20
Improve validation

dkfellows:
This does validation in py2json (instead of leaving its need noted as a `TODO` 🙄) so we should capture faults in configs earlier. It also fixes and simplifies some validators; Spring must be using custom validator loading code that worked before, but non-hosted validation is another use case.


rowleya (approved):
I think the numbers look OK; I would guess we will never reach them but we will find out if we do!

dkfellows:
I know that chip coords are limited, and that we can only operate by triads, and I think I worked out that this was as far as we can go. If it's wrong, we can just bump the limits up.

And we had them before anyway. This was just moving them around a bit.

PR #704, opened 2023-01-30
Improvements to Permit-handled auth tracking

dkfellows:
Apparently, sometimes the credentials internally used to authorize another (worker) thread to do an action on the user's behalf will leak. 

I've no idea how that happens! It's an extremely serious bug if it does; those tokens must not be handed from one thread to another at the wrong time or used outside of the issuing context. Right now, this is just about trying to make things easier to track down.


dkfellows:
I believe this was really about catching bugs elsewhere that have now been fixed,

PR #710, opened 2023-01-31
Add regular pinging on websocket

dkfellows:
nginx times out if you don't send a message from time to time (60 seconds?) but websockets support ping/pong frames as part of the base protocol. This starts using them.


dkfellows:
https://github.com/TooTallNate/Java-WebSocket/wiki/Lost-connection-detection

PR #712, opened 2023-02-02
ID issuing and tracking should be on a per-connection basis

dkfellows:
I'm not sure we can get away with a change this simple. Multi-threaded demuxing may also be required.


dkfellows:
It turns out that the demux could be handled by the `SCPConnection` provided I told it what the sequence number was. (I could have parsed the message to get that, but it's just as easy to get the code to pass that explicitly when sending a message that expects a reply.) That keeps this change a lot simpler than it would have otherwise been.

dkfellows:
The other thing about the demux is that even if the sender thread that wants a reply isn't waiting *right now,* it will be soon. And we can definitely optimize for the case where there's only one thread using connection; I expect that to be the norm for all non-root boards (and probably to normally be the case for roots as well).

PR #717, opened 2023-02-06
More type constraints on SCP messages

dkfellows:
Introducing `EmptyResponse`, which lets us more clearly distinguish between messages that produce payloads in their reponse, messages that have empty payloads, and messages that don't have responses at all. This matters because they have different usage patterns.

`IPTagGet` is a special case in that making it aware of what connection it is sent on allows us to let its response produce full `Tag` instances. I'd thought that more messages would care, but it seems this isn't the case at the moment, so what I have there is slightly over-engineered.


dkfellows:
Also moves model classes (such as `TagDescription`) to the `...model` package.

Issue #723, opened 2023-02-09
Change citation to ASL licensed


dkfellows:
The citation (`CITATION.cff`) is currently for version 6.0.0 of the software, which is _definitely_ GPL-3 licensed. Once (*and not until*) we have a citation for version 7, we should update the reference.

PR #728, opened 2023-02-14
License tweak a few more files

dkfellows:
I'd missed some places, either because they were RAT-excluded or because they were metadata. (Split out from #726 because that's too big a change.)

As noted in #723, `CITATION.cff` has not had its metdata section updated because 6.0.0 is released under the GPL. We will need to fix when 7 goes out, but can't until we have the DOI.


Christian-B (approved):
agreed.

PR #730, opened 2023-02-14
changed start date to match what git log say

Christian-B:
ran a script to do.
git log --diff-filter=A --follow --format=%aD -1

Where this came up with a year earlier than the copyright it changed the copyright year.

### DIFF
--- a/SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/board/ValidTriadWidth.java
+++ b/SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/board/ValidTriadWidth.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 The University of Manchester
+ * Copyright (c) 2021 The University of Manchester
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The University of Manchester
+# Copyright (c) 2018 The University of Manchester
 #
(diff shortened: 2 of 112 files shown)


SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/board/ValidTriadWidth.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 The University of Manchester
+ * Copyright (c) 2021 The University of Manchester
```
dkfellows, line 2: This is an actively wrong change. I only wrote it a few months ago, though there were other similar files prior to that (I'd been experimenting for about a week prior).

dkfellows:
Something has gone very wonky! I looked in detail at `ValidTriadWidth.java` (a file recent enough that I can remember what I was thinking when writing it) and it was definitely created in the first commit in the repo it appears in (with similar precursors about a week earlier). I don't know where the `2021` that you assign to it comes from, but it definitely isn't correct.

SpiNNaker-machine/src/main/java/uk/ac/manchester/spinnaker/machine/board/ValidTriadWidth.java
Christian-B: This is one of many where the Script took the add date of the file that was copied to create this one.

So when was the work actually started?

Issue is in many case I imagine the copy was just to get the copyright headers so that is not really work started.

Christian-B:
Suggest closing this PR as just using the git add date is too aggressive

Christian-B:
Done better in another PR.
Thanks @dkfellows or pointing out the error

PR #742, opened 2023-02-21
Board Temperature Reading

### LINKED ISSUE #720: Display temperatures in board UI
It would be nice if the web UI for looking at boards and/or the machine provided information about the temperatures of the boards (with caching for some amount of time). We have the call available to get that information from the BMPs; it would need to be exposed (and that means touching the BMP controller).

dkfellows:
This eventually fixes #720


dkfellows:
FWIW, 5f9ab73 has a strange label simply because I used the wrong commit comment.

PR #745, opened 2023-02-27
Tidy up now we use MySQL

dkfellows:
* Add code to get the result columns (except where that fails due to a [MySQL connector bug](https://bugs.mysql.com/bug.php?id=103437)...). This is intended for use in testing; it's not designed to be fast.
* Add code to get the list of parameters too. I've wanted that for a while! This is intended for use in testing; it's not designed to be fast.
* Augment the SQL-related tests to use those above operations where possible.
* Fix the declarations of some of the SQL queries so their annotations match reality.
* Use [`UNIX_TIMESTAMP()`](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_unix-timestamp) instead of passing `Instant.now()` explicitly; I think we can count on the DB knowing when is now.
* Clean up a few assignments to be of `final` fields or to use `var`. While I was in the area.


dkfellows:
Note that this will eventually add back in a lot of tests that were removed by #737 (relating to what the parameters and results of each DB query are). I believe they were removed because access to the metadata was... irritatingly tricky to fix. (It was one of those "need to do exactly the right thing" sort of challenges.)

rowleya (approved):
This all seems fine.  The failures in integration tests seem unrelated to these changes...

PR #753, opened 2023-03-02
Size of machine in compat mode is in chips

### LINKED ISSUE #743: Spalloc Job is not described in terms of chip width and height
The Spalloc job size for the compat service should return the size in terms of the width and height of the job in chips.  I believe this requires knowledge of the wrap-arounds of the machine, which is also not provided.

dkfellows:
This fixes #743

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -805,6 +805,10 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 
 		private final int height;
 
+		private final boolean hWrap;
+
+		private final boolean vWrap;
+
 		@JsonIgnore
 		private final Epoch epoch;
 
@@ -815,6 +819,8 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 			width = rs.getInt("width");
 			height = rs.getInt("height");
 			inService = rs.getBoolean("in_service");
+			hWrap = rs.getBoolean("horizontal_wrap");
+			vWrap = rs.getBoolean("vertical_wrap");
 			try (var getTags = conn.query(GET_TAGS)) {
 				tags = Row.stream(copy(getTags.call(string("tag"), id)))
 						.toSet();
@@ -1019,6 +1025,16 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 		@Override
 		public String toString() {
 			return "Machine(" + name + ")";
+		}
+
+		@Override
+		public boolean isHorizonallyWrapped() {
+			return hWrap;
+		}
+
+		@Override
+		public boolean isVerticallyWrapped() {
+			return vWrap;
 		}
 	}
 
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Spalloc.java
@@ -805,9 +805,11 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 
 		private final int height;
 
-		private final boolean hWrap;
+		private boolean lookedUpWraps;
 
-		private final boolean vWrap;
+		private boolean hWrap;
+
+		private boolean vWrap;
 
 		@JsonIgnore
 		private final Epoch epoch;
@@ -819,8 +821,7 @@ public class Spalloc extends DatabaseAwareBean implements SpallocAPI {
 			width = rs.getInt("width");
 			height = rs.getInt("height");
 			inService = rs.getBoolean("in_service");
-			hWrap = rs.getBoolean("horizontal_wrap");
-			vWrap = rs.getBoolean("vertical_wrap");
+			lookedUpWraps = false;
 			try (var getTags = conn.query(GET_TAGS)) {
 				tags = Row.stream(copy(getTags.call(string("tag"), id)))
 						.toSet();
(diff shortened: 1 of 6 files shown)


rowleya:
This doesn't appear to be correct, as it doesn't account for wrap arounds or non-wrap arounds in the system.  For example, a 3-board system from a 24-board system will not have wrap arounds, and the dimensions are therefore 16x16 (with holes but that is dealt with elsewhere), but this will return 12x12.  A 12-board allocation from a 24-board system will likely have wrap-arounds in a single dimension (similarly a 6-board allocation from a 12-board machine).

The calculation required is here:
https://github.com/SpiNNakerManchester/spalloc_server/blob/master/spalloc_server/coordinates.py#L265-L296

That requires knowledge of whether there are wrap-arounds or not, which I think comes from here:
https://github.com/SpiNNakerManchester/spalloc_server/blob/master/spalloc_server/allocator.py#L993-L1051

I don't know if this can be added in SQL form to get whether a job has wrap arounds or not...

dkfellows:
The width/height is the difference between the minimum coordinate and the maximum coordinate, yes? (Wraparounds won't affect that; we're in job-world coordinates already at this point which account for that, and whether the wraps matter for the allocation is irrelevant to the size.)

I've got to double-check a diagram...

rowleya:
The width and height is the width and height of the booted machine.  This is affected by wrap-arounds because in a wrap around system the top/right extents wrap around and fill in gaps in the bottom/left extents.  Without wrap-arounds those bits that *would* wrap around stick out instead.  So a 3 board system with wrap arounds is 12x12 but a 3 board system without is 16x16.

dkfellows:
DirInfo1

🤔 

OK, I need to add ~4~ `MachineDefaults.HALF_SIZE` to each coordinate after multiplying by ~12~ `MachineDefaults.TRIAD_WIDTH`/`MachineDefaults.TRIAD_HEIGHT`.

rowleya:
That is true *only* if you don't have all boards in one or other dimension i.e. it depends on what has been allocated.  The machine *will* have wrap-arounds if you allocate a certain set of boards.  If you allocate 12 boards from that machine as below, it will wrap around at the top/bottom.

222687913-6823a8bc-0bad-40bb-aee5-1bebc275d087

rowleya:
The machine below would wrap around only on the left-right.  I don't think this is actually a likely allocation on this machine, but it is a valid machine...
222687913-6823a8bc-0bad-40bb-aee5-1bebc275d087

dkfellows:
Yes, you wrap then, but that doesn't affect the width or height; in your first example, the height is still 28 whether or not the wraparound links exist. (I also don't know whether the code would pick the root to be 0,0,0 or 0,1,0; either would be valid. It might split between them due to the "wear levelling" of root nodes.)

rowleya:
> Yes, you wrap then, but that doesn't affect the width or height; in your first example, the height is still 28 whether or not the wraparound links exist. (I also don't know whether the code would pick the root to be 0,0,0 or 0,1,0; either would be valid. It might split between them due to the "wear levelling" of root nodes.)

That isn't the case.  In the first example, the height would be 24 because the top bit of 0, 1, 2 and 1, 1, 2 would actually fill in the gaps to the right of 0, 0, 0 and 1, 0, 0.

You could be right that the root board might end up being 0, 1, 0 if the boot code gets there first.  That doesn't really matter though; the machine dimensions will still be the same.

rowleya:
It looks a bit like this (poorly edited) diagram:

222687913-6823a8bc-0bad-40bb-aee5-1bebc275d087_wrap

dkfellows:
Did the traditional spalloc actually do that?

rowleya:
Yes, traditional spalloc calculated this correctly having looked at the actual machine wrap arounds.  Note that this is a consequence of the links in the machine; it will happen because of how SCAMP assigns addresses to chips and how the links are connected.  This isn't a spalloc decision, just an understanding of the connectivity of the resulting machine.

dkfellows:
Since the DB has the knowledge of the links (they're a full table) we can pick that info up in the queries that create the machine.

dkfellows:
Dang. No materialized views in MySQL...

dkfellows:
I'm not yet sure if I ought to preemptively compute whether the machine has wraps or not. It's fine to have it as a *concept* that the machine knows, and it definitely isn't going to change for the lifetime of a machine, but I'm not sure whether it's info that is sufficiently frequently needed to require computing on first creation of the `MachineImpl`. The (sub)queries required are a little bit complicated.

dkfellows:
That should work, assuming I haven't made a semantic blunder in the queries... but I don't know if it is the best way of doing it. Maybe the cost of discovering if there are wraparounds should be borne only when someone actually wants that information?

rowleya (approved):
I think this does the right thing now...

dkfellows:
The text of the new query (after re-wrapping for readability) is:
```sql
WITH linked AS (
    SELECT b1.machine_id, b1.x AS x1, b1.y AS y1, b2.x AS x2, b2.y AS y2 
    FROM links 
        JOIN boards AS b1 ON links.board_1 = b1.board_id
        JOIN boards AS b2 ON links.board_2 = b2.board_id)
SELECT 
    EXISTS (
        SELECT 1 FROM linked 
        WHERE linked.machine_id = machines.machine_id
            AND ((linked.x1 = 0 AND linked.x2 = machines.width - 1) 
            OR (linked.x2 = 0 AND linked.x1 = machines.width - 1)))
        AS horizontal_wrap,
    EXISTS (
        SELECT 1 FROM linked 
        WHERE linked.machine_id = machines.machine_id 
            AND ((linked.y1 = 0 AND linked.y2 = machines.height - 1) 
            OR (linked.y2 = 0 AND linked.y1 = machines.height - 1))) 
        AS vertical_wrap
FROM machines 
WHERE machine_id = :machine_id 
LIMIT 1
```

dkfellows:
The query will be technically wrong on an unwrapped Spalloc-controlled machine up to about 12 boards in size, but who cares? That's a very minority case.

dkfellows:
Suggestions for a better query are fine. They can go in their own PR.

PR #755, opened 2023-03-06
Fix logged errors in testing of compat code

dkfellows:
Don't know why there are empty strings being parsed (and failing) during the running of V1CompatTest, but this turns those into EOFs.


dkfellows:
One exception left in the compat code testing, and that I want to remain: it's thrown because we test with a deliberately wrong message, and we want it there (at level `ERROR`) because it will help any future debugging of clients required.

PR #760, opened 2023-03-06
Make some links secure

dkfellows:
Only the links where it actually matters have been changed


Christian-B (commented):
There are several more http which I am changing but this is all correct so approved.

Christian-B (approved):
There are several more http which I am changing but this is all correct so approved.

PR #763, opened 2023-03-08
Citation plus

Christian-B:
paart of https://github.com/SpiNNakerManchester/SpiNNMachine/pull/194


dkfellows:
Fixed the validation so that it is in the right place and enabled it.

And deleted an old irrelevant dependabot rule.

dkfellows (approved):

PR #765, opened 2023-03-08
Proxy protocol upgrade

dkfellows:
Allows some non-protocol errors to be reported in the proxy system. This includes asking for a connection to a board that is not allocated to you.


rowleya:
"Something" goes wrong on this - two runs of the code have resulted in the same error twice:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/proxy-upgrade/3
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/proxy-upgrade/4

dkfellows:
That's odd. That shouldn't be running via the proxy. It ought to be nothing to do with this branch.

rowleya:
The data gatherer has to run via the proxy if everything else is...  Note that the version without these changes passed recently:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/fix_proxy_listening/4/pipeline

rowleya:
The previous test run that was successful used the data gatherer via the proxy as far as I remember; in any case the previous test is the current master version that does work with the proxy settings here.  As far as I know, the only changes are those made in these branches, which seems to then fail...

dkfellows:
Well, I've adjusted the logging so that we'll get a little bit more info in the failure case (i.e., how much were we trying to write into how much space).

rowleya:
I have re-enabled synchronization on the session for sending on the session.  This appears to make it all work again...

dkfellows:
Well, [this SO question (and one of its answers)](https://stackoverflow.com/q/22257079/301832) leads to [Tomcat Issue#56026](https://bz.apache.org/bugzilla/show_bug.cgi?id=56026), and that makes me think this is a misinterpretation of the websocket spec by Tomcat, and one that they appear to want to be stubborn about. Also, Spring Boot doesn't appear to paper over this crack.

Synching on the session when sending appears to be the right approach, or at least as close to it as we'll get. The bug report has dark murmurings about it not being enough, but no suggested better option.

rowleya:
OK, since doing that has fixed the issue for us, I am going to merge this now

PR #787, opened 2023-04-18
Second attempt to fix the download reliability problem

dkfellows:
Maven was ignoring the setting I was using in #786. This other approach ought to work, provided the config is right. The documentation of this is very poor and confusing!

The only place this setting is documented is on [this page](https://maven.apache.org/resolver/configuration.html), and that says *nothing* about the correct way to convey the setting into the point that reads it. I've had to guess that it is a system property.


dkfellows:
That appears to have done the right thing; the `autobuild` step in the "**CodeQL / Analyze**" workflow was the most fragile one.

PR #884, opened 2023-06-15
Major cleanup/ reduction of the DataSpecification

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1069

PR #885, opened 2023-06-15
Error Prone doesn't support Java 14

dkfellows:
This isn't well documented *in the slightest,* but
https://github.com/google/guava/pull/6243 points to the right place (i.e., has useful information for people) and https://github.com/google/error-prone/issues/3540 is the core issue (i.e., it is the same stack trace that we've observed).

This just marks the error prone checking run as not to be done for 14. It already was only done if enabled by Maven profile.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,7 @@ jobs:
         cache: 'maven'
 
     - name: "Check: Error Prone"
+      if: matrix.java != 14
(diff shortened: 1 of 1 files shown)


Christian-B:
Needed for https://github.com/SpiNNakerManchester/JavaSpiNNaker/pull/884

Christian-B (approved):
this is a MUST!

dkfellows:
Hmm, https://checkerframework.org/manual/ *sort of* hints that it doesn't support Java 14... but doesn't spell it out at all. Argh!

PR #913, opened 2023-06-20
Fix the DB code

dkfellows:
This fixes a problem I noticed in #884, in that we were using the wrong query to count things. Positional binding of SQL is so easy to get wrong!

(It started as a private branch to scratch some itches relating to DB API usage; I didn't think it worth pushing yet. But updating the branch let me notice that some queries were wrong, and so it gets promoted to a critical bug fix.)


SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/sqlite/SQL.java
```diff
@@ -168,7 +171,21 @@ private SQL() {
...
 					+ " FROM core_view";
 
-	/** List the cores of a ethernets with a data specification to run. */
+	/**
+	 * Count of cores with data to load. This picks either system or application
+	 * cores.
+	 */
+	@Parameter("is_system")
+	@ResultColumn("num_cores")
+	static final String COUNT_CORES_TO_LOAD =
```
dkfellows, line 180: This query was outright missing.

SpiNNaker-storage/src/main/java/uk/ac/manchester/spinnaker/storage/sqlite/SQLiteDataSpecStorage.java
```diff
@@ -73,13 +74,12 @@ public int countCores(boolean loadSystemCores) throws StorageException {
 
 	private static int countCores(Connection conn, boolean loadSystemCores)
 			throws SQLException {
-		try (var s = conn.prepareStatement(LIST_CORES_TO_LOAD)) {
-			// ethernet_id
+		try (var s = conn.prepareStatement(COUNT_CORES_TO_LOAD)) {
```
dkfellows, line 72: This is where the wrong query was being used

Christian-B: agreed

Christian-B (approved):
looks good

PR #934, opened 2023-07-10
RemoteSpiNNaker

rowleya:
Adds RemoteSpiNNaker to JavaSpiNNaker and updates to include quota services

PR #935, opened 2023-07-10
RemoteSpiNNaker docker

rowleya:
This is the movement of RemoteSpiNNaker to JavaSpiNNaker, along with the addition of the use of Docker to launch virtual machines for Jobs.  This updates the Spalloc client a bit as well as it otherwise fails to work correctly with this service (which is the first time we have fully used the Java client I believe).


SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
```diff
@@ -163,7 +168,22 @@ public static Job getJobFromProxyInfo(ProxyInformation proxy)
...
+		String inputStr;
+		while ((inputStr = streamReader.readLine()) != null) {
+			responseStrBuilder.append(inputStr);
+		}
+		String json = responseStrBuilder.toString();
+
+		try {
+			return JSON_MAPPER.readValue(json, cls);
+		} catch (IOException e) {
+			log.error("Error while reading json " + json);
```
dkfellows, line 184: `json` isn't fully trusted at this point; log it with:
```java
log.error("Error while reading json {}", json);
```

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
```diff
@@ -354,11 +374,12 @@ private static final class ClientImpl extends Common
...
 
-		private ClientImpl(Session s, RootInfo ri) {
+		private ClientImpl(Session s, RootInfo ri) throws IOException {
 			super(null, s);
 			this.v = ri.version;
-			this.jobs = ri.jobsURI;
-			this.machines = ri.machinesURI;
+			this.jobs = asSecure(asDir(ri.jobsURI));
+			this.machines = asSecure(asDir(ri.machinesURI));
+			log.debug("Job URI = " + this.jobs);
```
dkfellows, line 382: Probably best to change this too.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
```diff
@@ -354,11 +374,12 @@ private static final class ClientImpl extends Common
 
 		private URI machines;
 
-		private ClientImpl(Session s, RootInfo ri) {
+		private ClientImpl(Session s, RootInfo ri) throws IOException {
 			super(null, s);
 			this.v = ri.version;
-			this.jobs = ri.jobsURI;
-			this.machines = ri.machinesURI;
+			this.jobs = asSecure(asDir(ri.jobsURI));
```
dkfellows, line 380: I wonder why the URI wasn't secure in the first place? Points to a different problem...

SpiNNaker-comms/pom.xml
```diff
@@ -71,6 +71,10 @@ limitations under the License.
 			<groupId>org.java-websocket</groupId>
 			<artifactId>Java-WebSocket</artifactId>
 		</dependency>
+		<dependency>
+		    <groupId>org.apache.httpcomponents</groupId>
+		    <artifactId>httpclient</artifactId>
```
dkfellows, line 76: Does the client use this dependency?

dkfellows (changes requested):
A few small things, mostly relating to care when logging (because anything that originates outside the service might include `{}` in its string, confusing the logger's template engine).

The HTTP/HTTPS thing is *probably* just a testing vs deployment issue. I hope!

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
rowleya: This is what the spalloc server returns.  Does this point to a configuration issue at the spalloc server?  It is running http internally but is behind an nginx https server...

dkfellows: It's *supposed* to trust the proxying headers put in by nginx (that's something that has to be enabled in config IIRC, which is fair enough).

rowleya: Ah yes, that was the issue - the URLs are now https after changing the nginx config on spalloc.  Maybe then this shouldn't be done...

dkfellows:
How does this relate to #934?

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
dkfellows: It's always best to ensure that URLs are correct to start out with instead of munging them client-side.

rowleya:
They are mostly the same, but this one is an extension of that one.  Given you have reviewed this one I will probably delete that one...

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
rowleya: Fixed

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
rowleya: Removed altogether as part of something else

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/alloc/client/SpallocClientFactory.java
rowleya: Removed now

SpiNNaker-comms/pom.xml
rowleya: Once the asSecure code is removed this is no longer used

dkfellows (approved):

PR #937, opened 2023-07-17
Fix bmp

rowleya:
Fixes a few things to make the BMP work.  The short vs. integer parts are questionable as to whether they are needed, but there was an issue with negative sequence numbers occurring, so the hope is that this could fix them.

### DIFF
--- a/SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/scp/SCPRequestHeader.java
+++ b/SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/scp/SCPRequestHeader.java
@@ -29,11 +29,14 @@ import uk.ac.manchester.spinnaker.messages.SerializableMessage;
  * actual sequence number when the message is sent on a connection.
  */
 public final class SCPRequestHeader implements SerializableMessage {
+
+	private static final int SHORT_MASK = 0xFFFF;
+
 	/** The command of the SCP packet. */
 	public final CommandCode command;
 
 	/** The sequence number of the packet, between 0 and 65535. */
-	private short sequence;
+	private int sequence;
 
 	private boolean sequenceSet;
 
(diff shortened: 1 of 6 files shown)


SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/scp/SCPRequestHeader.java
```diff
@@ -74,20 +77,20 @@ public void addToBuffer(ByteBuffer buffer) {
...
 	 *             If an attempt is made to set a sequence number a second time
 	 */
-	public short issueSequenceNumber(Set<Integer> inFlight) {
+	public int issueSequenceNumber(Set<Integer> inFlight) {
 		if (sequenceSet) {
 			throw new IllegalStateException(
 					"a message can only have its sequence number set once");
 		}
 		do {
 			sequence = getNextSequenceNumber();
```
dkfellows, line 86: You might want to use `Short.toUnsignedInt()` on this line, so we don't have negative sequence numbers floating around.

rowleya: I think this avoids the negative numbers because they are no longer short values until the last second (when they are sent) or when they are received (when I think the conversion is currently correct)

rowleya: Oh I see your point now - instead I will make that return an int too!

rowleya: I think this is now correct (because the method now returns an int as I thought it did before)

dkfellows:
I suspect you'd have never written this if I'd written the guts of the `BMPCommandProcess` to handle "casting" right, but this is probably more elegant anyway.

SpiNNaker-comms/src/main/java/uk/ac/manchester/spinnaker/messages/scp/SCPRequestHeader.java
dkfellows: OK

dkfellows (approved):

PR #938, opened 2023-07-17
We should only get jobs with the right permit

rowleya:


dkfellows:
This makes the jobs of non-v1 users impossible to see, as opposed to the current policy which lets people see them (and get some basic info, _not including ownership or connectivity!_) but otherwise not do very much. Is that what we want?

dkfellows:
> if it doesn't give an error that is better than now!

Fair enough.

PR #951, opened 2023-07-20
fix job boards

rowleya:
Attempt to fix issue where boards are left allocated to jobs

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/db/SQLQueries.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/db/SQLQueries.java
@@ -215,15 +215,13 @@ public abstract class SQLQueries {
 					+ "SELECT MAX(x) - MIN(x) + 1 AS width, "
 					+ "MAX(y) - MIN(y) + 1 AS height FROM c LIMIT 1";
 
-	/** Get what boards are allocated to a job (that is queued or ready). */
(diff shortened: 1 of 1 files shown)


rowleya:
The BMPController has been significantly rewritten now so that it includes a Thread per BMP, which is periodically run and gets tasks for each BMP.  The BMPController and Allocator also now have references to each other to ping when things change, but note this is just done to speed things up in some cases; it isn't critical that it runs in order to work.  Additionally, when a job is destroyed, the outstanding requests are not deleted; instead additional requests are added to make sure that transitions still occur, especially as there isn't really a global overview over all the BMPs.

rowleya:
Tested with http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/fix_job_boards/6/pipeline/147/

pom.xml
```diff
@@ -359,6 +359,11 @@ limitations under the License.
 			    <artifactId>mysql-connector-j</artifactId>
 			    <version>${mysql.version}</version>
 			</dependency>
+			<dependency>
+			    <groupId>org.hobsoft.spring</groupId>
+			    <artifactId>spring-rest-template-logger</artifactId>
```
dkfellows, line 364: Just to be sure, is this something present but usually disabled in config? We don't want to log requests normally.

rowleya: (It needs log level DEBUG to show the messages)

dkfellows (approved):

PR #992, opened 2023-08-11
Fix the notification of compat jobs

rowleya:
Most of this branch adds debug messages, but of important significance is the change of the Epochs so that the set doesn't get removed during an update.  This allows an epoch to be reused again, which is critical in the compat jobs where the same Epoch is used multiple times.  The weak map and periodic cleanup ensures that the map doesn't leak (verified through manual testing).


dkfellows:
Not working yet... 😕

dkfellows:
You might need to do significant work on this branch after #994.

rowleya:
This now appears to work correctly.

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/bmp/BMPController.java
```diff
@@ -156,6 +156,7 @@ private void init() {
 		var sched = new ThreadPoolTaskScheduler();
 		scheduler = sched;
 		sched.setThreadGroupName("BMP");
+		sched.initialize();
```
dkfellows, line 159: There's already a call to that, later in the method and _after_ the pool size has been set. I think this call is unnecessary.

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Epochs.java
```diff
@@ -264,21 +275,21 @@ class EpochMap {
...
 			for (var item : items) {
 				item.updateChanged(id);
 			}
 		}
 	}
 
-	private synchronized Set<Epochs.Epoch> removeSet(int id) {
-		return map.remove(id);
+	private synchronized Set<Epochs.Epoch> getSet(int id) {
+		return map.get(id);
```
dkfellows, line 292: I've been wondering whether this set should be copied so that the references become strong ones...

Something for another PR I believe.

dkfellows (changes requested):

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/Epochs.java
dkfellows: See #1028

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/bmp/BMPController.java
rowleya: That's what I thought too, but it turns out that makeWorkers was calling the scheduler before it was initialised then.  I have reordered the calls and this appears to fix things and not call initialize twice.

dkfellows (approved):

PR #1035, opened 2023-09-05
Retry proxy

rowleya:
Adds a proxy that can be used to help with server restarts

### DIFF
--- a/SpiNNaker-allocserv/pom.xml
+++ b/SpiNNaker-allocserv/pom.xml
@@ -186,27 +186,6 @@ limitations under the License.
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-jar-plugin</artifactId>
-				<executions>
-					<execution>
-						<phase>package</phase>
-						<goals>
-							<goal>jar</goal>
-						</goals>
-						<configuration>
-							<classifier>proxy</classifier>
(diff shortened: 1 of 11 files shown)


dkfellows:
I wonder whether the proxy code should be its own submodule.

dkfellows:
Also, how is this code used? 😉

rowleya:
At present, the servers are started with a non-normal compat port, and the proxy is then run against the real port connecting to this non-normal port (e.g. the proxy listens on 22244 and the server listens on 22254).  If we restart the server and leave the proxy running, the user will not be disconnected, and the should be able to continue using the software when the service resumes (normal restarts are quick enough that jobs won't expire).

rowleya:
It isn't super critical, it was just to allow me to restart things when users are busy :)

dkfellows:
So this will be something that gets deployed separately. Which means it needs to go in the deployment instructions, and probably should also be its own submodule (as it is its own tiny little artefact).

dkfellows:
FWIW, I approve of the principle. I'm nitpicking.

rowleya:
> So this will be something that gets deployed separately. Which means it needs to go in the deployment instructions, and probably should also be its own submodule (as it is its own tiny little artefact).

Yes to both those things; I will try to split it out

dkfellows (approved):

dkfellows:
I suppose it ought to be writing to a logfile instead of stderr. Fixable later.

PR #1065, opened 2023-09-18
Order connected boards correctly

dkfellows:
The query to discover the connected boards at a candidate root coordinate and list them in order (so the root comes first) was wrong; it was ordering by the global coordinates, yet that's wrong when the candidate allocation is wrapped. Need to order by the (candidate) job-local coordinates instead.

The upshot of this was that some branches of the tools became very confused as to what was the root board and didn't work as a result.


rowleya (approved):
Tested on demo server

PR #1101, opened 2023-10-20
Handle bmp errors

rowleya:
Ensures that BMP errors do not result in success!  Also ensures that BMPs are not powered on until at least 30 seconds after they have been powered off.  This does not affect allocations; the boards will be allocated but will wait off until the 30 seconds have passed.  This won't affect most users, just those who ask for specific boards (normal users will likely be allocated other boards anyway).

Also adds some features to the UI (power on/off machine and show real user on compatibility jobs).


rowleya:
Tests here are mostly passing (against the demo server with the changes above):
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/handle_bmp_errors/1/pipeline

PR #1134, opened 2024-01-19
Fix openid login

rowleya:
Revert some things to make it work, but still hopefully make it not check server when starting up by providing all the values needed.


rowleya:
Apologies, this also ended up fixing issues with the build and the test that have just appeared!

PR #1135, opened 2024-02-26
Job fail on alloc failed

rowleya:
Adds option to remove job queue altogether

### DIFF
--- a/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
+++ b/SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
@@ -387,9 +387,6 @@ public class AllocatorTask extends DatabaseAwareBean
 		/** Get the list of allocation tasks for jobs in a given state. */
 		private final Query getTasks;
 
-		/** Delete an allocation task. */
-		private final Update delete;
-
 		/** Find a single free board. */
 		private final Query findFreeBoard;
 
@@ -439,7 +436,6 @@ public class AllocatorTask extends DatabaseAwareBean
 			super(conn);
 			bumpImportance = conn.update(BUMP_IMPORTANCE);
 			getTasks = conn.query(getAllocationTasks);
-			delete = conn.update(DELETE_TASK);
 			findFreeBoard = conn.query(FIND_FREE_BOARD);
 			getRectangles = conn.query(findRectangle);
 			getRectangleAt = conn.query(findRectangleAt);
@@ -455,7 +451,6 @@ public class AllocatorTask extends DatabaseAwareBean
 			super.close();
 			bumpImportance.close();
 			getTasks.close();
-			delete.close();
 			findFreeBoard.close();
 			getRectangles.close();
 			getRectangleAt.close();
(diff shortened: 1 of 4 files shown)


SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
```diff
@@ -651,13 +646,18 @@ private Allocations allocate(Connection conn) {
 					continue;
 				}
 				var handled = task.allocate(sql);
-				// If we handled it, delete the request
-				if (handled.size() > 0) {
-					sql.delete.call(task.id);
```
Christian-B, line 656: I think you still need this line as otherwise the job stays in requested forever

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
```diff
@@ -651,13 +646,18 @@ private Allocations allocate(Connection conn) {
...
+
+				if (handled.size() == 0
+						&& allocProps.isFailJobOnFailAllocation()) {
```
Christian-B, line 651: Longer term this needs to be fancier but ok for now

Christian-B (changes requested):

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
rowleya: I think this was an error introduced previously.  I thought that we needed to remove the job queue item when it was allocated but actually a lot of things can still go wrong between a job being allocated and a job being ready meaning that you need to leave it on the queue.  Note that the job queue table is not the same as the job queue.  The job queue is the combination of job entries and job queue entries that are in the queued state; once allocated a job goes into powering, which is either successful so it goes to ready, or fails, so it goes back in the queue...

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
rowleya: We can make this fancier now; the failure or otherwise of a job can be more fine grained if desired and then allow for better queuing...

SpiNNaker-allocserv/src/main/java/uk/ac/manchester/spinnaker/alloc/allocator/AllocatorTask.java
Christian-B: Yes I see how src/main/resources/queries/get_allocation_tasks.sql can handle things staying in the job_request table forever.

If however we want to switch to keeping the job permanently in both the jobs and job_request tables it raises the question of should this be two tables or just one as there will now be a permanent  one to one mapping between the two tables.

Christian-B: Not calling sql.delete.call(task.id) will help with a job where the allocation fails temporarily as stated.
This avoids the danger of that single job remaining status QUEUE forever

But if whatever caused the allocation to fail will happen very time the task will increase in importance until it blocks the Queue for all other tasks!

So then it becaomes a choise of which evil is preferred.

Christian-B:
Approved to merge Assuming this passed all tests.

In a follow up pr we can look at which jobs to keep in the queue if no boards available.

Christian-B (approved):
my comments remain but are not blockers

rowleya:
This has been done elsewhere now.

PR #1137, opened 2024-04-17
add abort on timeout

Christian-B:


Christian-B:
This was tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/abort/1/pipeline/
Which got past the Java build crashing because spin4 pinged but would not turn on.

That then failed for other reasons which may point out why spin4 is misbehaving

PR #1223, opened 2025-04-10
Fix swagger

rowleya:
Allows the "swagger" docs to be visible and makes sure it shows our API!


rowleya:
This is *almost* working.  The changes here do work for all except the base URI for some reason...

Christian-B (approved):

PR #1281, opened 2025-08-20
Bump more versions

rowleya:
Fixes issues with tests, and bumps versions as requested by dependabot


rowleya:
Note, a load of fields of private classes have been made explicitly private where possible, as this makes checkstyle no longer complain about them (due to checkstyle update)

PR #1407, opened 2026-01-05
Java 21 25

rowleya:
A number of updated dependencies are starting to move away from Java 17 to minimum Java 21.  We already support Java 21, but this would remove support for 17 and introduce support for the latest Java LTS version 25.  I think we are already using Java 21 everywhere that matters.


Christian-B (commented):
please check you did mean 2,24 and not 2.25

